PR #7, opened 2019-01-22
Acceptance tests

tbarabosch:
Implemented with pytest for convenience. Run full test suite with make test.


Enkelmann:
Right now all tests are failing on my computer, because no warnings are generated at all.

PR #8, opened 2019-02-04
Add dockcross for C++ x86 cross compilation.

Enkelmann:
We could also use dockcross containers for all other cross compiling targets if we want to.


test/artificial_samples/install_cross_compilers_ubuntu_18.04.sh
```diff
@@ -5,4 +5,6 @@ echo "Installing cross compiler for MIPS architecture."
 sudo apt install -y gcc-multilib-mips-linux-gnu g++-7-mips-linux-gnu
```
tbarabosch: can't we cross compile everything with dockcross? By doing so, we would not need to pollute our systems with the cross compiling toolchains of arm, mips, and so on.

tbarabosch: if this is correct, then we can remove the other dependencies.

test/artificial_samples/makefile
```diff
@@ -36,7 +36,7 @@ endef
```
tbarabosch: can't we change everything to dockcross?

test/artificial_samples/dockcross_x86
```diff
@@ -0,0 +1,254 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
Enkelmann: was auto-generated from dockcross, now the install script generates it.

Enkelmann:
I have replaced all cross compiling toolchains with dockcross. The downside is the higher disk space required (about 4GB alltogether, can probably be optimized if we care about it).

test/artificial_samples/makefile
Enkelmann: Done.

test/artificial_samples/install_cross_compilers_ubuntu_18.04.sh
Enkelmann: Done.

PR #9, opened 2019-02-05
Simple recipe for standard run

Enkelmann:
A simple recipe for a standard run using the internal rooter of bap.


tbarabosch:
Looks fine!

PR #13, opened 2019-02-21
Pre commit 

tbarabosch:
Initial pre-commit support. Will add ocaml checks later.


src/config.json
```diff
@@ -1,58 +1,180 @@
...
+      "stpncpy",
+      "wcpcpy",
+      "wcpncpy",
+      "memcpy",
+      "wmemcpy",
+      "memmove",
+      "wmemmove",
+      "memcmp",
+      "wmemcmp",
+      "me\u200bmset",
```
Enkelmann: Shouldn't this be memset without \u200b?

src/config.json
```diff
@@ -1,58 +1,180 @@
...
+      "wmemset",
+      "gets",
+      "sprintf\u200b",
```
Enkelmann: Shouldn't this be sprintf without \u200b?

src/config.json
tbarabosch: Some encoding issues, fixed by now.

PR #14, opened 2019-03-29
Type inference

Enkelmann:
Contains pass that checks which registers hold pointers and which hold data other than pointers. Also tracks values on the stack. Stack access may get reworked when adding pointer target tracking, i.e. tracking to what heap object/function stack frame a pointer is pointing to.

Oh, also contains some unit tests. :-)


tbarabosch:
First of all: it already looks very nice! :+1: 

Some general issues:
- please add a recipe for type-inference-printer, would be a really nice to have!
- same behavior of dune as stated by you: some changes require now make clean && make
- in order to make it work with BAP's opam testing, I had to fix some Map.adds, they are now Map.add_exn (grep for them, but there should be also TODO comments), wrapped up in try-with-blocks. Please have a look at them, just in case that I fixed something. It would be even better if you could preserve the Map.adds and unpack them properly (OK a | Duplicate). 
- regarding the general Makefile: if you run make test it fails because the C samples havent been built yet. We should build them before. Also there should be a make option: "make samples". This could be done in another story.

tbarabosch:
Once again, let's switch now to BAP testing in this branch, this implies using Core 0.11.0
FYI:
opam repo add bap git://github.com/BinaryAnalysisPlatform/opam-repository#testing
opam update
opam install depext --yes
opam depext --install bap --yes

src/utils/symbol_utils.ml
```diff
@@ -147,3 +147,17 @@ let get_program_entry_points program =
   let entry_points = Seq.filter subfunctions ~f:(fun subfn -> Term.has_attr subfn Sub.entry_point) in
   let main_fn = Seq.filter subfunctions ~f:(fun subfn -> "@main" = Tid.name (Term.tid subfn)) in
   Seq.append main_fn entry_points
+
```
tbarabosch: These three functions should not be here. The module Symbol_utils deals with imported/exported ELF symbols.
Since these functions are all machine related, there could be a module Machine_utils or something similar.

src/utils/address_translation.ml
```diff
@@ -11,5 +11,5 @@ let generate_tid_map prog =
     inherit [addr Tid.Map.t] Term.visitor
     method enter_term _ t addrs = match Term.get_attr t address with
       | None -> addrs
-      | Some addr -> Map.add addrs ~key:(Term.tid t) ~data:addr
+      | Some addr -> Map.add_exn addrs ~key:(Term.tid t) ~data:addr
```
tbarabosch: I believe this should never blow up, otherwise the project is very broken anyways.

src/checkers/cwe_476.ml
```diff
@@ -273,7 +273,7 @@ let check_cwe prog proj tid_map symbol_names parameters =
   Seq.iter subfunctions ~f:(fun subfn ->
       let cfg = Sub.to_cfg subfn in
       let cwe_hits = ref [] in
-      let empty = Map.empty Graphs.Ir.Node.comparator in
+      let empty = Map.empty (module Graphs.Ir.Node) in
```
tbarabosch: Core 0.11.0 changed some things. You do not need to hand over a comparator, just hand over a module.

src/analysis/type_inference.mli
```diff
@@ -0,0 +1,45 @@
...
+    then the output format should be refactored accordingly. *)
+val print_type_info_tags: project:Project.t -> tid_map:word Tid.Map.t -> unit
+
+(** Updates the type info for a single element (Phi/Def/Jmp) of a block. Input
+    is the type info before execution of the element, output is the type info
+    after execution of the element. *)
+val update_type_info: Blk.elt -> TypeInfo.t -> project:Project.t -> TypeInfo.t
+
+(* functions made public for unit tests: *)
+val update_block_analysis: Blk.t -> TypeInfo.t -> project:Project.t -> TypeInfo.t
```
tbarabosch: Is that idiomatic? What does Alcotest recommend?

src/utils/cconv.ml
```diff
@@ -0,0 +1,84 @@
...
+
+let is_callee_saved var project =
+  match !callee_saved_registers with
+  | Some(register_set) -> String.Set.mem register_set (Var.name var)
+  | None ->
+    callee_saved_registers := Some(String.Set.of_list (callee_saved_register_list project));
+    String.Set.mem (Option.value_exn !callee_saved_registers) (Var.name var)
+
+(** Parse a line from the dyn-syms output table of readelf. Return the name of a symbol if the symbol is an extern function name. *)
+let parse_dyn_sym_line line =
```
tbarabosch: This function could be in Symbol_utils

src/utils/cconv.ml
```diff
@@ -0,0 +1,84 @@
...
+  match !str_list with
+  | _ :: value :: _ :: "FUNC" :: _ :: _ :: _ :: name :: [] -> begin
+      match ( String.strip ~drop:(fun x -> x = '0') value, String.lsplit2 name ~on:'@') with
+      | ("", Some(left, _)) -> Some(left)
+      | ("", None) -> Some(name)
+      | _ -> None (* The symbol has a nonzero value, so we assume that it is not an extern function symbol. *)
+    end
+  | _ -> None
+
+let parse_dyn_syms project =
```
tbarabosch: This function could be in module Symbol_utils

src/utils/cconv.ml
```diff
@@ -0,0 +1,84 @@
...
+  | None ->
+    callee_saved_registers := Some(String.Set.of_list (callee_saved_register_list project));
+    String.Set.mem (Option.value_exn !callee_saved_registers) (Var.name var)
+
+(** Parse a line from the dyn-syms output table of readelf. Return the name of a symbol if the symbol is an extern function name. *)
+let parse_dyn_sym_line line =
+  let line = ref (String.strip line) in
+  let str_list = ref [] in
+  while Option.is_some (String.rsplit2 !line ~on:' ') do
+    let (left, right) = Option.value_exn (String.rsplit2 !line ~on:' ') in
```
tbarabosch: That is a general concern and should be applied to the whole pull request (!): there are many Option.value_exn. I think they are never caught when they blow up. To honor the type system, we should unpack them correctly.

test/unit/analysis/mem_region_test.ml
```diff
@@ -0,0 +1,98 @@
...
+
+let check msg x = Alcotest.(check bool) msg true x
+
+let test_add () : unit =
+  let bv num = Bitvector.of_int num ~width:32 in
+  let x = Mem_region.empty () in
+  let x = Mem_region.add x "Five" ~pos:(bv 3) ~size:(bv 5) in
+  let x = Mem_region.add x "Seven" ~pos:(bv 9) ~size:(bv 7) in
+  let x = Mem_region.add x "Three" ~pos:(bv 0) ~size:(bv 3) in
+  check "add_ok" (Some(Ok("Five", bv 5)) = (Mem_region.get x (bv 3)));
```
tbarabosch: Some(Ok()) and Some(Error()), I am really not sure what to think about it. Yet I've never seen it in other OCaml projects.

test/unit/analysis/mem_region_test.ml
```diff
@@ -0,0 +1,98 @@
...
+
+let test_minus () =
+  let bv num = Bitvector.of_int num ~width:32 in
+  let x = Mem_region.empty () in
+  let x = Mem_region.add x "One" ~pos:(bv (-8)) ~size:(bv 8) in
+  check "negative_index" (Some(Ok("One", bv 8)) = Mem_region.get x (Bitvector.unsigned (bv (-8))))
+
+let test_remove () =
+  let bv num = Bitvector.of_int num ~width:32 in
+  let x = Mem_region.empty () in
```
tbarabosch: these let x = look were tedious. Cann't you use the |> operator??

src/analysis/type_inference.ml
```diff
@@ -0,0 +1,454 @@
...
+            try
+              { state with TypeInfo.reg = Map.add_exn state.TypeInfo.reg addr (Ok(Register.Pointer)) } (* TODO: there are some false positives here for indices in global data arrays, where the immediate is the pointer. Maybe remove all cases with potential false positives? *)
+            with _ -> state
+            end
+          | Bil.BinOp(Bil.PLUS, Bil.Var(addr), exp2)
+          | Bil.BinOp(Bil.MINUS, Bil.Var(addr), exp2)
+          | Bil.BinOp(Bil.AND, Bil.Var(addr), exp2)
+          | Bil.BinOp(Bil.OR, Bil.Var(addr), exp2) ->
+            if type_of_exp exp2 state project = Some(Ok(Register.Data)) then
+              try
```
tbarabosch: Did I break something?

tbarabosch:
This pull request is HUGE, let's start with the first couple of comments. I will do another iteration once they are fixed.

tbarabosch:
One more thing: we should prefix EVERY plugin with cwe_checker, i.e. cwe_checker_type_inference for instance!

tbarabosch:
And another thing: please adjust the Dockerfile to build everything (all the new plugins) for Travis CI!
Adjust the .travis* scripts to run Alcotest etc.

Enkelmann:
- When we switch to opam-testing for the master branch, we have to document it in the Readme and adjust the docker container.
- I changed the Map.add_exn to Map.set, because in these use cases the previous value is meant to be overwrittten if it exists.
- I will check whether prefixing every pass with cwe_checker works.

plugins/cwe_type_inference_print/cwe_type_inference_print.ml
```diff
@@ -0,0 +1,14 @@
+open Bap.Std
```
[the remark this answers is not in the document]
Enkelmann: As far as I can see, the -d option prints only terms, not attributes of terms. Also, it does not look like the output of -d is meant to be modified by plugins. So I suppose it would be easier to implement our own version of -d. This should not be too hard, but would still wait with that until type inference is feature complete.

Enkelmann:
Regarding the `Some(Ok(_))`-wrappings: I thought about defining a new wrapper type, but I think this way is more idiomatic: Every information during a fixpoint computation has essentially three states:
- We don't have the information -> `None`
- We do have conflicting information (e.g. from a merge) -> `Some(Error())`. We could also set this to None, but this could lead to infinite loops during fixpoint computation and knowing that there is conflicting infomation could also be useful information
- We do have non-conflicting information -> `Some(Ok(information))`

Unless you think we should define a new wrapper type for this I don't know a more idiomatic way.

Regarding Codacy: Didn't we already fix that space problem in changes.md? Did codacy change its mind?

src/analysis/mem_region.ml
```diff
@@ -0,0 +1,164 @@
...
+  let () = if pos + size < pos then failwith "[CWE-checker] element out of bounds for mem_region" in
+  let new_node = {
+    pos=pos;
+    size=size;
+    data=Ok(elem);
+  } in
+  match mem_region with
+  | [] -> new_node :: []
+  | head :: tail ->
+    if head.pos + head.size <= pos then
```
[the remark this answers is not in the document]
Enkelmann: I couldn't think of a useful way to extract a function there. I added some comments to improve readability instead.

test/unit/analysis/mem_region_test.ml
Enkelmann: I cannot use |> here, because x is not the last argument in the Mem_region.add calls. I could write an add_list function for Mem_region for nicer looking code. But I actually like the current version more, because this way it is very explicit in which order the elements get added.

test/unit/analysis/mem_region_test.ml
Enkelmann: See https://github.com/fkie-cad/cwe_checker/pull/14#issuecomment-478556193

Enkelmann:
TODO: Update readme for use of core 0.11

src/analysis/type_inference.mli
Enkelmann: The documentation of alcotest says nothing about this. l think it may be more idiomatic to wrap such functions into a Test submodule. I will do that for the time being, unless we find an even more idiomatic way.

src/utils/cconv.ml
Enkelmann: Hmm, in this particular case I would like to keep the value_exn, since I checked in the previous line that it cannot throw an exception. There is a way of writing this without the value_exn using functional programming style, but I think this would be more verbose and less readable in this case.

PR #15, opened 2019-03-29
cwe_checker_emulation plugin using BAP's Primus

tbarabosch:
This pull request adds
- cwe_checker_emulation plugin based on BAP's Primus. It detects CWE-125/CWE-787, CWE-415, CWE-416
- Some fixes due to changes in JaneStreets Core library (Hashtbl, Map)
- It is now based on BAP's opam testing branch -> this may break stuff, we have to upgrade the Travis image as well


tbarabosch:
I propose that we first resolve #14  and merge it into this PR. 
Then we can use cwe_checker_core and get rid of some duplicate code in cwe_checker_emulation (Log_utils).

plugins/cwe_checker_emulation/cwe_checker_emulation.ml
```diff
@@ -0,0 +1,178 @@
...
+- all monitored events are collected globally
+- after the last Primus machine has terminated we report all observed incidents *)
+let main {Config.get=(!)} proj =
+  Log_utils.set_log_level Log_utils.DEBUG;
+  Log_utils.set_output stdout;
+  Log_utils.color_on ();
+
+  Primus.Machine.add_component (module Monitor);
+  begin
+  let targets =  all_subroutines (Project.program proj) in
```
Enkelmann: Why are entry points filtered out from targets?

plugins/cwe_checker_emulation/cwe_checker_emulation.ml
```diff
@@ -0,0 +1,178 @@
...
+    else entries
+  end
+
+(** Wrapper function around entry_point_collector. *)
+let entry_points prog =
+  entry_point_collector#run prog []
+
+(** Collects all subroutines of the program that
+that are not an entry point. *)
+let all_subroutines prog =
```
Enkelmann: I don't really like the name `all_subroutines`, since entry points are filtered out.

plugins/cwe_checker_emulation/cwe_checker_emulation.ml
```diff
@@ -0,0 +1,178 @@
...
+let entry_point_collector = object
+  inherit [tid list] Term.visitor
+  method! enter_term _ t entries =
+    if Term.has_attr t Sub.entry_point
+    then Term.tid t :: entries
+    else entries
+  end
+
+(** Wrapper function around entry_point_collector. *)
+let entry_points prog =
```
Enkelmann: There is a similar function in symbol_utils, except it also adds the main function as entry point. I assume adding the main function here would break something? This function should probably be moved to symbol_utils.

plugins/cwe_checker_emulation/incident_reporter.ml
```diff
@@ -0,0 +1,56 @@
...
+let report incident location_tbl =
+  match incident with
+  | name::ids ->
+    begin
+      let incident_locations = get_incident_locations_from_ids ids location_tbl in
+      let filtered_locs = Int.Set.to_list (Int.Set.of_list (List.concat incident_locations)) in
+      let incident_str = Sexp.to_string name in
+      match incident_str with
+      (* TODO: we report an out-of-bounds read but actually it could also be
+         an out-of-bound write. Find out how to distinguish between them. *)
```
Enkelmann: Then I would suggest that we report an out-of-bounds read/write for the time being.

tbarabosch: I am going to report an incident for each of them. Above all, this solution makes the parsing later much more easier.

plugins/cwe_checker_emulation/cwe_checker_emulation.ml
tbarabosch: I am using a function from Symbol_utils now.

plugins/cwe_checker_emulation/cwe_checker_emulation.ml
tbarabosch: I renamed this function to collect_non_entry_subroutines.

plugins/cwe_checker_emulation/cwe_checker_emulation.ml
tbarabosch: The basic idea is to focus on a per subroutine emulation.

tbarabosch: Now we try to emulate all subroutines.

tbarabosch:
I added tests for cwe-415 and cwe-416. 
Unfortunately, we are not able to find a working example for cwe-125/cwe-787. There are two artificial examples that should work but I skipped the tests.

PR #16, opened 2019-03-29
Build artificial samples with scons instead of make

tbarabosch:
This PR exchanges make with scons. This makes building the artificial samples painless!!!


test/artificial_samples/SConstruct
```diff
@@ -0,0 +1,72 @@
...
+    if shutil.which(c_compilers[arch]) is not None:
+        c_programs = Glob('*.c')
+        for p in c_programs:
+            if compile_only_on_x64(str(p), arch):
+                print('Skipping architecture %s for %s' % (arch, str(p)))
+                continue
+            env = Environment(CC = c_compilers[arch],
+                          CCFLAGS = c_flags[arch] + optimize(str(p)))
+            env.Program('build/%s_%s.out' % (str(p).split('.')[0], arch), 'build/%s' % str(p))
+            # TODO: call execstack -s
```
Enkelmann: Do we have a test right now that depends on the call of execstack?

tbarabosch: I believe none, I do not know why it was there in the first place.

tbarabosch:
Scons3 does not work properly on Ubuntu 16.04/18.04, but on current Fedora. So switching to scons2 (pip install --user scons). Now it should compile all binaries.

PR #20, opened 2019-06-18
Corrected dune linter warnings, linter warnings are now handled like errors.

Enkelmann:
The code in the plugins folder is not built by dune, thus it is also not checked by dune. For every other piece of Ocaml code the build process will now fail if dune emits a warning.


tbarabosch:
Looks fine, we're getting closer to clean code :)

PR #21, opened 2019-06-19
Cwe 560

tbarabosch:


src/checkers/cwe_560.mli
```diff
@@ -0,0 +1,21 @@
...
+version considers all immediate values of an umask callsite's basic block.
+*)
+val name : string
+val version : string
+
+val check_cwe : Bap.Std.program Bap.Std.term -> Bap.Std.project -> Bap.Std.word Bap.Std.Tid.Map.t -> string list list -> string list -> unit
+
+
+(* IMPORTANT: Do not use the functions below; they are just exported for testing with AlcoTest *)
+val is_chmod_style_arg : int -> bool
```
Enkelmann: As a convention, functions that are only public for unit tests should be put into a submodule named `Private`.

PR #27, opened 2019-07-18
Adds clang as second compiler for test cases

tbarabosch:
Even though clang is the first example, we could add more compilers in the future (e.g. the official ARM compilers).


test/acceptance/test_cwe476.py
```diff
@@ -8,29 +8,35 @@ def setUp(self):
...
         self.assertEqual(res, expect_res)
 
-    def test_cwe476_01_x64(self):
+    def test_cwe476_01_x64_gcc(self):
         expect_res = 1
-        res = cwe_checker_testlib.execute_and_check_occurence(self.target, self.target, 'x64', self.string)
+        res = cwe_checker_testlib.execute_and_check_occurence(self.target, self.target, 'x64', 'gcc', self.string)
+        self.assertEqual(res, expect_res)
+
+    @unittest.skip("FIXME")
```
Enkelmann: I should take a look at this.

Enkelmann: clang seems to optimze away the relevant calls for the check. Is there a "do not optimize anything"-flag for clang that we could use?

tbarabosch: we could build the binaries with -o0

tbarabosch:
I am going to look into the x86_gcc issue. But I suppose it's a problem with dockcross....

test/acceptance/test_cwe476.py
Enkelmann: If it is not enabled by default, we should do that.

tbarabosch:
I did a clean build of all artificial examples on this branch. I was able to build for all arches including x86. Could you please check your dockcross installation?

Enkelmann:
OK, then it is probably a problem on my machine. Merging.

Issue #34, opened 2019-07-30
The script for IDA doesn't work


dukebarman:
Generated script has got only two strings:

```
import sark
import idaapi
```

I think after refactoring of logging the output hasn't got "WARN" string so `cwe_checker_to_ida.py` cannot normal generate an IDA annotate script.

tbarabosch:
Oh yes, I am going to fix it ASAP.

tbarabosch:
PR #35 should fix this issue

dukebarman:
Yeap, it works. Thanks!

I think better to add info at readme that the IDA script for converting uses python3.

P.S. `Done. Now execute generated script ~/ida_script.py with IDAPython (alt+F9).` maybe `(alt+F7)`?

tbarabosch:
Yes, @Enkelmann is going to update the documentation.

PR #35, opened 2019-07-31
Fix cwe_checker_to_ida, which was broken due to refactoring of logging

tbarabosch:
see issue #34


cwe_checker_to_ida/cwe_checker_to_ida.py
```diff
@@ -24,9 +24,15 @@ def save_generated_script(outpath, generated_script):
 
 def main():
     args = parse_args()
+
+    if not args.cwe_checker_result.endswith('.json'):
```
Enkelmann: Can you check for a valid json file instead?

PR #40, opened 2019-11-09
Cwe caller

mellowCS:


src/caller/helper_functions.ml
```diff
@@ -0,0 +1,60 @@
...
+    match String.is_prefix ~prefix:element head with
+    | true  -> replacement::tail
+    | false -> head::replace_element tail element replacement
+
+
+(* Just for type conversion: Returns an empty string if none, else the string *)
+let get_default_string (in_option : string option) : string =
+  match in_option with
+  | None   -> ""
+  | Some s -> s
```
Enkelmann: There is already a method for Options which does the same.

src/caller/helper_functions.mli
```diff
@@ -0,0 +1,20 @@
...
+(* Just for type conversion: Returns an empty string if none, else the string *)
+val get_default_string: string option -> string
+
+(* Returns all elements only present in set_a *)
+val get_difference: 'a list -> 'a list -> 'a list
+
+(* Returns user input from the command line from given position as a string list *)
+val get_user_input: ?position:int -> unit -> string list
+
+(* Returns flag with file attachment if present *)
```
Enkelmann: This function is also used to find the `-partial` flag, so the documentation is slightly wrong. ;-)

Enkelmann:
The command line flags and parameters are now defined in the Cwe_checker_core.Main module. @mellowCS Could you refactor your code to read them from there? We should also print a custom help message for the *--help* flag.

Enkelmann:
Correction: We do want to keep the value checks for parameters here. My mistake. I will move them afterwards to the Main module so that we can use the same checks here and in the Main module.

caller/cwe_checker.ml
```diff
@@ -0,0 +1,169 @@
...
+  match String.is_suffix path ~suffix:"/" with
+  | true  -> generate_output_file path ~file:("out-" ^ string_of_float (Unix.time ())) ()
+  | false -> generate_output_file path ()
+
+
+let out_path (path : string) : string =
+  try
+    match Sys.is_directory path with
+    | false -> generate_output_file path ~file:"" ()
+    | true  -> build_path path
```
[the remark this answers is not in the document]
mellowCS: Can the user pass non-existing folders as well? If not, I would check if the given path is valid.

Enkelmann:
Looks good. I will add some refactoring of the main module and after that we can merge it.

PR #53, opened 2020-03-18
Test compiling

mellowCS:


mellowCS:
Have to check why the three tests pass on my system but not on travis.

Enkelmann:
The cwe416_01_x64_mingw_gcc fail looks like it could be related to changes in BAP Primus. That means it could depend on different snapshots of the BAP master branch on your system and in the docker builds. We probably should skip it until BAP 2.1 is out and revisit the issue then.

mellowCS:
So we are already running on BAP 2.0? If so, we may want to change that in the readme.

Enkelmann:
The master branch runs on a snapshot of the current master branch of BAP. The latest stable release still runs on BAP 1.6. But yes, the readme needs a general overhaul anyway.

mellowCS:
Updated from Ocaml 4.05.0 to 4.07.1  and recompiled everything but nothing has changed. I will skip the other tests and take a note so we can update them when we have moved to BAP 2.1.

PR #58, opened 2020-04-22
Symbol checking

mellowCS:


mellowCS:
Is it wise to merge into master without a running travis?

src/utils/symbol_utils.ml
```diff
@@ -8,18 +8,76 @@ type symbol =
...
+
+
+let get_project_calling_convention (project : Project.t) : string option =
+  Project.get project Bap_abi.name
+
+
+let build_extern_symbols (project : Project.t) (program : program term) : unit =
+  let parsed_symbols = String.Set.to_list (Cconv.parse_dyn_syms project) in
+  let calling_convention = get_project_calling_convention project in
+  let tid_map = Address_translation.generate_tid_map program in
```
Enkelmann, line 31: Right now generating the TID map is computationally expensive. It should be a parameter to the function, so that we only have to compute the map once.

src/utils/symbol_utils.ml
```diff
@@ -8,18 +8,76 @@ type symbol =
...
+    | false -> None)))
+
+
+let build_and_return_extern_symbols (project : Project.t) (program : program term) : extern_symbol list =
+  if List.is_empty !extern_symbols then
+    begin
+      build_extern_symbols project program;
+      !extern_symbols
+    end
+  else !extern_symbols
```
Enkelmann, line 50: In the case where there are no extern symbols, the symbol list would be built over and over again.

src/utils/symbol_utils.ml
```diff
@@ -8,18 +8,76 @@ type symbol =
...
+
+
+let add_extern_symbol (project : Project.t) (program : program term) (symbol : string) : unit =
+  let tid_map = Address_translation.generate_tid_map program in
```
Enkelmann, line 54: The TID map should be a parameter to the function.

src/utils/symbol_utils.mli
```diff
@@ -14,6 +14,28 @@ type symbol = {
...
+val get_project_calling_convention : Bap.Std.Project.t -> string option
+
+(** Checks whether the external symbols have already been built. If not, it calls the symbol builder. *)
+val build_and_return_extern_symbols : Bap.Std.Project.t -> Bap.Std.program Bap.Std.term -> extern_symbol list
+
+(** Builds a list of function symbols type from external function names given by objdump. *)
+val build_extern_symbols : Bap.Std.Project.t -> Bap.Std.program Bap.Std.term -> unit
+
+(** Adds an analysed internal symbol to the list of external symbols. *)
+val add_extern_symbol : Bap.Std.Project.t -> Bap.Std.program Bap.Std.term -> string -> unit
```
Enkelmann, line 37: I think the function name is slightly misleading. Maybe rename it in `add_as_extern_symbol` or `mark_as_extern_symbol`

caller/cwe_checker.ml
```diff
@@ -142,22 +140,37 @@ let process_input () : string * string list =
...
+      let params = List.map cmdline_params ~f:(fun param -> match param with | (p, _) -> "-" ^ p) in
+      check_params params input_params;
+      (binary_path, input_params @ process_flags flags)
     )
 
 
+let setup_command (bin_path : string) (args : string list) : string =
+  let api = List.exists args ~f:(fun arg -> String.is_prefix arg ~prefix:"-api") in
+  let bare_command = match api with
+  | true -> "bap " ^ bin_path ^ " --pass=cwe-checker,api "
```
Enkelmann, line 156: Is mentioning `api` here actually needed? Since it is now a dependency of the cwe_checker plugin, the api plugin should run anyway.

mellowCS: I don't know if I'll be able to call the --api-path param otherwise

src/utils/symbol_utils.ml
mellowCS: I'll change the code for the tid map generator. I've to see where it is used and then probably change it in some other files to.

PR #59, opened 2020-05-13
Adapt cconv

mellowCS:


src/utils/cconv.ml
```diff
@@ -6,93 +6,111 @@ let dyn_syms = ref None
...
-  | `ppc (* 32bit PowerPC *) (*  TODO: add floating point registers. *) (* TODO: add CR2, CR3, CR4. Test their representation in bap first. *)
-  | `ppc64 | `ppc64le -> (* 64bit PowerPC *)
-    "R14" :: "R15" :: "R16" :: "R17" :: "R18" :: "R19" :: "R20" :: "R21" :: "R22" :: "R23" ::
-    "R24" :: "R25" :: "R26" :: "R27" :: "R28" :: "R29" :: "R30" :: "R31" :: "R1" :: "R2" :: []
-  | `mips | `mips64 | `mips64el | `mipsel -> (* S8 and FP are the same register. bap uses FP, S8 is left there just in case. *)
-    "S0" :: "S1" :: "S2" :: "S3" :: "S4" :: "S5" :: "S6" :: "S7" :: "S8" :: "GP" :: "SP" :: "FP" :: []
-  | _ -> failwith "No calling convention implemented for the given architecture."
+let bin_format = ref ""
+
+let json (() : unit) : Yojson.Basic.t =
```
Enkelmann, line 11: We should save the result of this function in memory to prevent parsing the json file from disk every time this function gets called.

src/utils/json_utils.mli
```diff
@@ -3,3 +3,7 @@
 val get_symbol_lists_from_json : Yojson.Basic.t -> string -> string list list
 val get_symbols_from_json : Yojson.Basic.t -> string -> string list
 val get_parameter_list_from_json : Yojson.Basic.t -> string -> string list
+val get_arch_from_json : Yojson.Basic.t -> ?conv:string -> string -> Yojson.Basic.t 
+val get_bin_format_from_json : Yojson.Basic.t -> string -> Yojson.Basic.t
+val get_registers_from_json : Yojson.Basic.t -> string -> string list
+val get_arch_list_from_json : Yojson.Basic.t -> string -> (string * Yojson.Basic.t) list
```
Enkelmann, line 9: Could you add documentation strings to these functions?

test/unit/Makefile
```diff
@@ -3,6 +3,7 @@ all:
 	bapbuild -r -Is analysis,checkers,utils cwe_checker_unit_tests.plugin -pkgs alcotest,yojson,unix,ppx_jane,cwe_checker_core
 	bapbundle install cwe_checker_unit_tests.plugin
 	bap ../artificial_samples/build/arrays_x64_gcc.out --pass=cwe-checker-unit-tests
+	bap ../artificial_samples/build/arrays_x64_clang.out --pass=cwe-checker-unit-tests
```
Enkelmann, line 6: This causes *all* unit tests to run twice. I don't think that this is a good idea.

mellowCS: So we just run them with one file as before?

Enkelmann: That would be my suggestion. Most of the current unit tests do not depend on the file and for those that do we have to find a better solution anyway.

PR #62, opened 2020-06-03
Address translation

mellowCS:


src/utils/address_translation.ml
```diff
@@ -2,6 +2,9 @@ open Core_kernel
 open Bap.Std
 
 
+let last_addr = ref None
```
Enkelmann, line 5: `last_addr` should be a local variable of `generate_tid_map`, not a global variable.

test/unit/utils/address_translation_test.mli
```diff
@@ -0,0 +1,7 @@
+open Bap.Std
+open Core_kernel
+
+
+val example_project: Project.t option ref
```
Enkelmann, line 5: Since you don't use the example project in the unit test, could you remove it here?

PR #83, opened 2020-09-03
Fixed a bug in the objdump output parsing in symbol_utils.ml

mellowCS:


src/utils/symbol_utils.ml
```diff
@@ -36,26 +36,12 @@ let call_finder_run = ref false
...
-          else None
-        end
-      | _ -> None (* The symbol has a nonzero value, so we assume that it is not an extern function symbol. *)
-    end
-  | _ -> None
+  let columns = String.split_on_chars ~on:[ ' ' ; '\t' ; '\n' ; '\r' ] line
+                |> List.filter ~f:(fun x -> x <> "") in
+  (* Check whether the symbol is a function --> DF and if it is referenced in the file, but defined outside it --> *UND* *)
+  match ((Stdlib.List.mem "DF" columns) && (Stdlib.List.mem "*UND*" columns)) with
+  | true -> Some(Array.last (Array.of_list columns))
```
Enkelmann, line 43: Why not use `List.last` instead of converting to an array first?

Enkelmann (changes requested):
PRs that simplify code are always the best PRs! :-)

Enkelmann (approved):

PR #86, opened 2020-10-12
Added p_code_extractor to cwe checker

mellowCS:


ghidra/p_code_extractor/PcodeExtractor.java
```diff
@@ -0,0 +1,1091 @@
...
+     * 
+     * Set register name based on being a register, virtual register, constant or ram address.
+     * In case it is a virtual register, prefix the name with $U.
+     * In case it is a constant, remove the const prefix from the constant.
+     */
+    protected Variable createVariable(Varnode node, Boolean isArgument) {
+        Variable var = new Variable();
+        if (node.isRegister()) {
+            String mnemonic = getRegisterMnemonic(node);
+            if(isArgument && mnemonic.equals("AL")){
```
Enkelmann, line 881: We cannot check for specific register names here, since this is not architecture-independent! Also, I would move the handling of this special case into its own function that only gets called when handling external symbols.

ghidra/p_code_extractor/PcodeExtractor.java
```diff
@@ -0,0 +1,1091 @@
...
+
+    /**
+     * 
+     * @param var: register variable
+     * @param node: varnode containing half register
+     * @return: full register variable
+     * 
+     * Casts half registers to full registers
+     */
+    protected Variable castToFullRegister(Variable var, Varnode node) {
```
Enkelmann, line 914: I think there is a function in the Ghidra API that does this independent of the register and independent of the CPU architecture.

ghidra/p_code_extractor/PcodeExtractor.java
```diff
@@ -0,0 +1,1091 @@
...
+            String callString = ghidraProgram.getLanguage().getUserDefinedOpName((int) pcodeOp.getInput(0).getOffset());
+            return new Call(createLabel(mnemonic, pcodeOp, null), createLabel(mnemonic, pcodeOp, instr.getFallThrough()), callString);
+        }
+        return new Call(createLabel(mnemonic, pcodeOp, null), createLabel(mnemonic, pcodeOp, instr.getFallThrough()));
+    }
+
+
+    /**
+     * @param address: Virtual register address
+     * @return: Prefixed virtual register naem
```
Enkelmann, line 1045: Typo

Enkelmann (changes requested):
We should split up the 1000 lines long file into several files. The refactoring necessary to reduce the number of parameters for some functions may be a good starting point for that.

ghidra/p_code_extractor/PcodeExtractor.java
```diff
@@ -0,0 +1,1091 @@
...
+
+    /**
+     * 
+     * @param sym: external symbol
+     * @return: if same symbol name in references
+     * 
+     * Checks whether the same symbol name is in the references of the current symbol.
+     * If so, the current symbol is not internally called by other functions
+     */
+    protected Boolean notInReferences(Symbol sym) {
```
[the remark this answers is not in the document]
mellowCS: Often there is a chain of external symbols like: some_function() -> system() -> system() -> external_system(). Using notInReferences I check whether the current symbol is the function (e.g. system()) in the middle of the chain or not. If it is than the same symbol name is going to be in the fromReferences of the middle symbol. I only want the symbol that is called by some_function()

ghidra/p_code_extractor/PcodeExtractor.java
```diff
@@ -0,0 +1,1091 @@
...
+     * @param blocks: array pf block terms
+     * @param instruction: assembly instruction
+     * @param pcodeOp: pcode instruction
+     * @param mnemonic: pcode mnemonic
+     * @param temporaryDefStorage: temporarily stored definitions
+     * @param instructionIndex: index of the current assembly instruction
+     * @param numberOfInstructionsInBlock: number of assembly instructions in Ghidra generated block
+     * @param numberOfPcodeOps: number of pcode instruction in pcode block
+     * @param pcodeIndex: index of current pcode instruction
+     * @param intraInstructionJumpOccured: indicator whether a jump occured inside a pcode block
```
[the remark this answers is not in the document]
mellowCS: I'll do that

ghidra/p_code_extractor/PcodeExtractor.java
```diff
@@ -0,0 +1,1091 @@
...
+     * @param blocks: array of block terms 
+     * @param instruction: assembly instruction
+     * @param instructionIndex: index of the assembly instruction
+     * @param numberOfInstructionsInBlock: number of assembly instructions in Ghidra generated block
+     * 
+     * Checks whether the assembly instruction is a nop instruction and adds a jump to the block.
+     * Checks whether a jump occured within a ghidra generated pcode block and fixes the control flow
+     * by adding missing jumps between artificially generated blocks.
+     */
+    protected void processPcode(ArrayList<Term<Blk>> blocks, Instruction instruction, int instructionIndex, long numberOfInstructionsInBlock) {
```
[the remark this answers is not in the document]
mellowCS: ok

ghidra/p_code_extractor/PcodeExtractor.java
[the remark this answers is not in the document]
Enkelmann: Ah, makes sense. Could you add this explanation to the doc comment?

ghidra/p_code_extractor/PcodeExtractor.java
mellowCS: Done

ghidra/p_code_extractor/PcodeExtractor.java
mellowCS: Done

ghidra/p_code_extractor/PcodeExtractor.java
mellowCS: Done

ghidra/p_code_extractor/PcodeExtractor.java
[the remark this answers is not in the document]
mellowCS: I will

Enkelmann (changes requested):
There is still a small unresolved comment from my first review. But it looks we are getting close to being able to merge it. :-)

Enkelmann (approved):

PR #88, opened 2020-10-22
Cwe 676 rs

mellowCS:


cwe_checker_rs/src/checkers/cwe_676.rs
```diff
@@ -0,0 +1,108 @@
...
+pub static CWE_MODULE: crate::CweModule = crate::CweModule {
+    name: "CWE676",
+    version: VERSION,
+    run: check_cwe,
+};
+
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Hash, Clone)]
+pub struct Config {
+    symbols: Vec<String>,
+}
```
Enkelmann, line 19: Could you add doc strings to this struct?

cwe_checker_rs/src/checkers/cwe_676.rs
```diff
@@ -0,0 +1,108 @@
+/// The CWE 676 checks for potentially dangerous function calls, such as strcpy, memcpy, scanf etc.
```
Enkelmann, line 1: The doc comment here should begin with `//!` (since it is inside the module which it documents) and not `///` (which needs to be outside of the object it documents. Also: We should insert the lengthy documentation for the CWE that we have in the corresponding `.mli` right now.

cwe_checker_rs/src/checkers/cwe_676.rs
```diff
@@ -0,0 +1,108 @@
...
+};
+
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Hash, Clone)]
+pub struct Config {
+    symbols: Vec<String>,
+}
+
+
+/// For each block of a subroutine check if the jump is a direct call and if the target TID of that call is equal to a dangerous symbol TID
+pub fn get_call_to_target<'a>(caller: &'a Term<Sub>, target: &'a ExternSymbol) -> Vec<Option<(&'a str, &'a Tid, &'a str)>> {
```
Enkelmann, line 23: You can remove the `Option` in the return type, since you only add the CWE hits to the Vec anyway.

cwe_checker_rs/src/checkers/cwe_676.rs
```diff
@@ -0,0 +1,108 @@
...
+    let mut calls: Vec<Option<(&'a str, &'a Tid, &'a str)>> = Vec::new();
+    for blk in caller.term.blocks.iter() {
+        for jmp in blk.term.jmps.iter() {
+            match &jmp.term {
+                Jmp::Call{ target: dst, .. } => {
+                    if *dst == target.tid {
+                        calls.push(Some((caller.term.name.as_str(), &blk.tid, target.name.as_str())));
```
Enkelmann, line 30: You should use the `jmp.tid` instead of the `blk.tid` here. That we needed to use block TIDs was a restriction we do not have anymore.

cwe_checker_rs/src/checkers/cwe_676.rs
```diff
@@ -0,0 +1,108 @@
...
+
+
+/// For each subroutine and each found dangerous symbol, check for calls to the corresponding symbol
+pub fn get_calls_to_symbols<'a>(subfunctions: &'a Vec<Term<Sub>>, dangerous_symbols: &'a Vec<&'a ExternSymbol>) -> Vec<Option<(&'a str, &'a Tid, &'a str)>> {
+    let mut calls: Vec<Option<(&str, &Tid, &str)>> = Vec::new();
+    for sub in subfunctions.iter() {
+        for symbol in dangerous_symbols.iter() {
+            calls.append(&mut get_call_to_target(sub, symbol))
+        }
+    }
```
Enkelmann, line 49: If you put the dangerous symbols into an appropiate HashMap or HashSet, we would not need the inner loop, which in turn would speed up the theoretical runtime of the algorithm. I don't know why we did not do this on the Ocaml side...

cwe_checker_rs/src/utils/log.rs
```diff
@@ -12,6 +12,44 @@ pub struct CweWarning {
...
+    pub fn symbols(mut self, symbols: Vec<String>) -> CweWarning {
+        self.symbols = symbols;
+        self
+    } 
+
+    pub fn other(mut self, other: Vec<Vec<String>>) -> CweWarning {
+        self.other = other;
+        self
+    } 
+}
```
Enkelmann, line 51: To be extra pedantic: Could you add doc strings to these functions?

Enkelmann (changes requested):
This already looks quite good! In addition to the comments below we need to run `cargo fmt` on the code to appease the CI.

cwe_checker_rs/src/checkers/cwe_676.rs
mellowCS: Added

Enkelmann (changes requested):
There are some doc strings missing in `cwe_782.rs` and `symbol_utils.rs`.

I am looking into the remaining unit test error, it should originate in my code and not in your code.

Edit: The unit test should be fixed now.

Enkelmann (approved):

PR #89, opened 2020-10-27
Doc image

mellowCS:


Enkelmann (changes requested):
The images look very good!
Right now the links in the rendered documentation look a little like regular in-doc-links, so the reader might not expect to find an explanatory image behind them. I think it will be clearer if we use something like `... [(image)](link/to/image) ...` for the link. Just to make it obvious that an image will be found when clicking the link.
The link text could be even more descriptive, like `(example graph image)`. But I am not sure what the best text would be.

Enkelmann (approved):

PR #91, opened 2020-10-28
add Ghidra-based acceptance checks to CI pipeline

Enkelmann:


test/src/lib.rs
```diff
@@ -0,0 +1,266 @@
...
+        mark_architecture_skipped(&mut tests, "mips"); // TODO: Check reason for failure!
+        mark_architecture_skipped(&mut tests, "mipsel"); // TODO: Check reason for failure!
+        mark_architecture_skipped(&mut tests, "ppc64"); // TODO: Check reason for failure!
+        mark_architecture_skipped(&mut tests, "ppc64le"); // TODO: Check reason for failure!
+        mark_architecture_skipped(&mut tests, "ppc"); // TODO: Check reason for failure!
+        mark_skipped(&mut tests, "x86", "gcc"); // TODO: Check reason for failure!
+        mark_compiler_skipped(&mut tests, "mingw32-gcc"); // TODO: Check reason for failure!
+
+        for test_case in tests {
+            if let Err(error) = test_case.run_test("[CWE415]", 2) {
```
mellowCS, line 188: Do you think it would be sensible to assigned the number to a variable like `let expected_occurrences: u8 = 2;`? It says it at the function header of run_test but I have forgotten what the number stood for when I came across the call and had to check the header again.

Enkelmann: Sounds like a good idea. Implemented.

mellowCS:
LGTM!

PR #94, opened 2020-10-29
Substitute trivial expressions

Enkelmann:
This PR adds a normalization pass to the `Project` struct that substitutes trivially computable expressions like `a XOR a` with their result.

The PR should only be merged after the **Fix implicit stores** PR is merged.

### DIFF
--- a/cwe_checker_rs/src/intermediate_representation/expression.rs
+++ b/cwe_checker_rs/src/intermediate_representation/expression.rs
@@ -76,6 +76,62 @@ impl Expression {
             Cast { size, .. } | Unknown { size, .. } | Subpiece { size, .. } => *size,
         }
     }
+
+    /// Substitute some trivial expressions with their result.
+    /// E.g. substitute `a XOR a` with zero or substitute `a OR a` with `a`.
+    pub fn substitute_trivial_operations(&mut self) {
+        use Expression::*;
+        match self {
+            Var(_) | Const(_) | Unknown { .. } => (),
+            Subpiece {
+                low_byte,
+                size,
+                arg,
+            } => {
+                arg.substitute_trivial_operations();
+                if *low_byte == ByteSize::new(0) && *size == arg.bytesize() {
+                    *self = (**arg).clone();
+                }
+            }
+            Cast { op, size, arg } => {
+                arg.substitute_trivial_operations();
+                if (*op == CastOpType::IntSExt || *op == CastOpType::IntZExt)
+                    && *size == arg.bytesize()
+                {
+                    *self = (**arg).clone();
+                }
+            }
+            UnOp { op: _, arg } => arg.substitute_trivial_operations(),
+            BinOp { op, lhs, rhs } => {
+                lhs.substitute_trivial_operations();
+                rhs.substitute_trivial_operations();
+                if lhs == rhs {
+                    match op {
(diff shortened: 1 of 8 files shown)


cwe_checker_rs/src/intermediate_representation/expression.rs
```diff
@@ -76,6 +76,62 @@ impl Expression {
...
+        use Expression::*;
+        match self {
+            Var(_) | Const(_) | Unknown { .. } => (),
+            Subpiece {
+                low_byte,
+                size,
+                arg,
+            } => {
+                arg.substitute_trivial_operations();
+                if *low_byte == ByteSize::new(0) && *size == arg.bytesize() {
```
mellowCS, line 92: Just for my understanding. You'll simplify the subpiece operation when the size of the subpiece is 0? So you basically remove subpiece und just have the underlying expression?

cwe_checker_rs/src/pcode/expressions.rs
```diff
@@ -122,6 +122,35 @@ pub struct Expression {
     pub input2: Option<Variable>,
 }
 
+impl From<Expression> for IrExpression {
+    /// Translates a P-Code expression into an expression of the internally used IR if possible.
+    /// Panics if translation is not possible,
+    /// e.g. for `LOAD`, `STORE` and and expressions that need the size of the output variable to be defined.
```
mellowCS, line 128: I think there is an expression type missing

cwe_checker_rs/src/intermediate_representation/expression.rs
Enkelmann: No, the size of the subpiece is the same as the size of the argument. And the subpiece starts at the zeroth byte of the argument (`low_byte = 0`). Or in other words, the subpiece is exactly the same as the argument.

cwe_checker_rs/src/pcode/expressions.rs
Enkelmann: You are right. I will clarify the doc comment!

test/src/lib.rs
```diff
@@ -178,9 +178,8 @@ mod tests {
         mark_architecture_skipped(&mut tests, "mips64el"); // TODO: Check reason for failure!
         mark_architecture_skipped(&mut tests, "mips"); // TODO: Check reason for failure!
         mark_architecture_skipped(&mut tests, "mipsel"); // TODO: Check reason for failure!
-        mark_architecture_skipped(&mut tests, "ppc64"); // TODO: Check reason for failure!
-        mark_architecture_skipped(&mut tests, "ppc64le"); // TODO: Check reason for failure!
-        mark_architecture_skipped(&mut tests, "ppc"); // TODO: Check reason for failure!
+        mark_architecture_skipped(&mut tests, "ppc64"); // Ghidra generates mangled function names here for some reason.
+        mark_architecture_skipped(&mut tests, "ppc64le"); // Ghidra generates mangled function names here for some reason.
```
mellowCS, line 182: So we don't skip 'ppc' anymore?

Enkelmann: Yes. :-) The problem with 'ppc' was a trivial expression of the form `a OR a`, that the analysis could not handle. By substituting it away, the check now works for 'ppc'.

PR #95, opened 2020-11-02
Cwe560

Enkelmann:
This PR reimplements the check for CWE 560 in Rust. It also is a slight improvement over the previous check, since we now actually make sure to only check the parameter value of umask and no other constants.


mellowCS (approved):
LGTM!

PR #96, opened 2020-11-02
Indirect call flows

mellowCS:


ghidra/p_code_extractor/PcodeExtractor.java
```diff
@@ -998,14 +1020,10 @@ protected Label handleLabelsForIndirectCalls(PcodeOp pcodeOp) {
...
-            if(referenced.length != 0) {
-                for (ExternSymbol symbol : program.getTerm().getExternSymbols()) {
-                    if (symbol.getAddress().equals(referenced[0].getToAddress().toString())) {
-                        return symbol.getTid();
-                    }
-                }
+        Address[] flowDestinations = PcodeBlockData.instruction.getFlows();
+        for(Address flow : flowDestinations) {
+            if(functionEntryPoints.containsKey(flow.toString())){
+                return new Tid(String.format("sub_%s", flow.toString()), flow.toString());
```
Enkelmann, line 1026: I think we will have a problem with false positives here. If a function has more than one flow, then it may jump to different addresses (that is if I interpret the Ghidra docs correctly). But we only want to detect cases where there is a unique call target here, since we use the information to replace indirect calls with direct calls. So we have to make sure that there are no other flows here!

If this should not work as intended (i.e. if we cannot reliably only detect those cases with a unique jump target), we will have to scrap this approach and replace it with a sound analysis on the Rust side.

ghidra/p_code_extractor/PcodeExtractor.java
```diff
@@ -90,6 +93,25 @@ protected void run() throws Exception {
...
+     * This will later speed up the cast of indirect Calls.
+     */
+    protected void setFunctionEntryPoints() {
+        // Add external symbols and internal function addresses to hash map
+        int funcCounter = 0;
+        for(ExternSymbol sym : program.getTerm().getExternSymbols()){
+            functionEntryPoints.put(sym.getAddress(), funcCounter);
+            funcCounter++;
+        }
+        funcCounter++;
```
Enkelmann, line 107: Has this line any practical meaning or can it be removed?

Enkelmann (changes requested):

Enkelmann (approved):

PR #101, opened 2020-11-04
Fix pointer to external function

mellowCS:
This PR adds multiple addresses to external symbols in case they are thunked. In case only a pointer to the external function exists, the external address is used as the sub TID. In case of jumps to function pointers of external symbols, the pointer address is added to the symbol's addresses. Furthermore, the addresses field now replaces the address field on the Rust side.


Enkelmann (approved):
LGTM

PR #103, opened 2020-11-09
retarget nonexisting jump targets to dummy target

Enkelmann:
This PR adds an artificial sink as a target for jumps to nonexisting TIDs. Such jumps may be generated by the Ghidra backend. It is deliberately not already corrected in the Ghidra script, since solving it later via an artificial sink has the upshot that we can generate an error log message for the user whenever such a jump is encountered.


mellowCS (approved):
LGTM!

PR #107, opened 2020-11-16
Generic indirect call behaviour

Enkelmann:
This PR adds CallStub edges for indirect calls. We also implement a standard behavior in the Pointer Inference analysis for indirect calls where we know nothing about the call target address. This allows continuation of the Pointer Inference analysis in case of unresolved indirect calls (albeit with lower analysis accuracy).


mellowCS (approved):
LGTM

PR #108, opened 2020-11-18
Add fall through branch to cbranch

mellowCS:
- Fallthrough branches are now added after CBRANCH instructions with their target depending on whether the CBRANCH occured inside of a pcode block or at the end.
- The currently processed pcodeOp and the pcodeIndex now reside in the PcodeBlockData struct, so that the index can properly
be handled from anywhere.
- Further refactorings were done for jump related functions


Enkelmann (changes requested):

ghidra/p_code_extractor/PcodeExtractor.java
```diff
@@ -368,38 +368,35 @@ protected void addMissingJumpAfterInstructionSplit(Term<Blk> lastBlock) {
...
     }
 
 
     /**
      * 
      * @param mnemonic: pcode mnemonic
      * @param numberOfPcodeOps: number of pcode instruction in pcode block
-     * @param intraInstructionJumpOccured: indicator whether a jump occured inside a pcode block
      * @param currentBlock: current block term
      * @return: indicator whether a jump occured inside a pcode block
```
Enkelmann, line 394: Your removed the return value from the function.

Enkelmann (changes requested):

Enkelmann (approved):

PR #112, opened 2020-12-02
Assume standard callee saved

Enkelmann:
This PR changes the behavior of the pointer inference analysis:
We now assume that all internal function calls adhere to the standard calling convention of the corresponding architecture. This allows us to restore the contents of callee-saved registers from the the caller state on a return instruction. Note that this new assumption should work reasonably well for programs written in C, but may introduce errors for programs that also use other calling conventions for internal function calls.


cwe_checker_rs/src/analysis/fixpoint.rs
```diff
@@ -83,13 +83,17 @@ pub trait Context {
...
+    /// The context object needed for the fixpoint computation
     fp_context: T,
     /// maps a node index to its priority (higher priority nodes get stabilized first)
     node_priority_list: Vec<usize>,
     /// maps a priority to the corresponding node index
     priority_to_node_list: Vec<NodeIndex>,
-    worklist: BinaryHeap<usize>,
+    /// The worklist contains the priority numbers (not the node indices!) of nodes marked as not yet stabilized.
+    worklist: BTreeSet<usize>,
+    /// The (optional) default value assigned to nodes without an explizit value.
```
mellowCS, line 94: change 'explizit' to 'explicit' :)

mellowCS (changes requested):
LGTM. Just one little typo :D

PR #113, opened 2020-12-02
more exact address reporting for CWE 676

Enkelmann:
Report the exact address of the call and not only of the basic block containing the call for CWE 676 (with BAP backend).


mellowCS (approved):
LGTM

PR #115, opened 2020-12-08
Cwe476 parameter checks

Enkelmann:
This PR adjusts how CWE hits are detected on function calls and function returns to reduce false positives.

For pointers to the stack as parameters, we underapproximate the values that the call may access to reduce false positives.
For pointers to the heap as parameters, we flag if any possible NULL pointer is contained in the target memory object.

### DIFF
--- a/cwe_checker_rs/src/checkers/cwe_476/state.rs
+++ b/cwe_checker_rs/src/checkers/cwe_476/state.rs
@@ -5,7 +5,7 @@ use crate::analysis::pointer_inference::Data;
 use crate::analysis::pointer_inference::State as PointerInferenceState;
 use crate::intermediate_representation::*;
 use crate::prelude::*;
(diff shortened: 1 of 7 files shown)


cwe_checker_rs/src/checkers/cwe_476/state.rs
```diff
@@ -222,25 +222,114 @@ impl State {
...
+                if elem.is_tainted() {
+                    return true;
+                }
+            }
+        }
+        false
+    }
+
+    /// If the given address points to the stack,
+    /// return true if and only if the value at that tsack position is tainted.
```
mellowCS, line 238: Small typo at the word 'stack'

Enkelmann:
Since all tests worked locally on my computer, I will merge it. We probably have to wait for Travis a few more days...

PR #116, opened 2020-12-10
Add the POPCOUNT  expression

Enkelmann:
This PR adds support for the POPCOUNT P-Code instruction introduced in Ghidra 9.2.

As a side effect, we are now fully compatible to Ghidra 9.2 and since Ghidra now includes the Gson library, one does not need to manually add it to Ghidra when installing the cwe_checker.


mellowCS (approved):
LGTM

PR #117, opened 2020-12-10
CWE 367

Enkelmann:
This PR reimplements the check for CWE 367 (Time-of-check-time-of-use race condition) in Rust for the Ghidra backend.

The `POPCOUNT` Pull Request should be merged before merging this PR !

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,10 +53,9 @@ Just run `make all` to compile and register the plugin with BAP. You can run the
 
 ### Local installation with Ghidra as backend ###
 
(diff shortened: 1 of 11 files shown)


mellowCS (commented):
Somehow your POPCOUNT PR is also in this one. Is it because it is needed for the check?

Enkelmann:
No, the POPCOUNT PR is here because I wanted to test the PR with Ghidra 9.2. In other words, I was lazy. ;-)

PR #118, opened 2020-12-16
Direction independent graph

mellowCS:
This PR adds a CallSource Node, which is the counter part to the CallReturn Node, to the CFG. This will enable the merge of data flowing backwards through a call to the callsite.


cwe_checker_rs/src/analysis/graph.rs
```diff
@@ -242,14 +256,39 @@ impl<'a> GraphBuilder<'a> {
...
+                            self.return_addresses
+                                .entry(target.clone())
+                                .and_modify(|vec| vec.push((cs_node, return_node)))
+                                .or_insert_with(|| vec![(cs_node, return_node)]);
+                        } else {
+                            self.return_addresses
+                                .entry(target.clone())
+                                .and_modify(|vec| vec.push((source, return_node)))
+                                .or_insert_with(|| vec![(source, return_node)]);
+                        }
```
Enkelmann, line 291: I am pretty sure we do not want to do this. We use the `return_addresses` to add `CallReturn` nodes later on. Adding this case would mean we allow `CallReturn` nodes without corresponding `CallSource` nodes.  This would be an unwanted artifact in the control flow graph.

cwe_checker_rs/src/analysis/graph.rs
```diff
@@ -311,6 +350,7 @@ impl<'a> GraphBuilder<'a> {
         for (call_node, return_to_node) in self.return_addresses[&return_from_sub.tid].iter() {
             let (call_block, caller_sub) = match self.graph[*call_node] {
                 Node::BlkEnd(block, sub) => (block, sub),
+                Node::CallSource { source, .. } => source,
```
Enkelmann, line 353: As mentioned, the graph should now disallow `CallReturn` nodes without corresponding `CallSource` nodes. So the `Node::BlkEnd` case should panic.

src/config.json
```diff
@@ -1,4 +1,27 @@
...
+      "scanf",
+      "__isoc99_scanf"
+    ],
+    "check_character_symbols": [
+      "strchr",
+      "strrchr",
+      "regexp",
+      "regexec"
+    ]
+  },
```
Enkelmann, line 24: Wrong Pull Request. ;-)

Enkelmann (changes requested):
We also need to update the nice graph pictures for the documentation you created, since the old pictures will be outdated with this PR.

cwe_checker_rs/src/analysis/graph.rs
```diff
@@ -52,9 +52,11 @@ pub type Graph<'a> = DiGraph<Node<'a>, Edge<'a>>;
 /// The node type of an interprocedural control flow graph
 ///
 /// Each node carries a pointer to its associated block with it.
-/// For `CallReturn`nodes the associated blocks are both the callsite block (containing the call instruction)
+/// For `CallReturn`nodes the associated nodes are both the `CallSource`node (containing the call instruction)
 /// and the returning-from block (containing the return instruction).
 ///
+/// For `CallSource`nodes the associated node is the callsite block (source)
+///
```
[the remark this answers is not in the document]
mellowCS: But the CallSource Node does not represent a single Block. At least not in the backwards analysis.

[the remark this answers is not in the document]
Enkelmann: Yes, there are two blocks associated with it, just as with the `CallReturn` node. If my original formulation did not make this clear, feel free to change it. :-) But these are still blocks that are associated with the nodes, not other nodes.

Enkelmann (approved):
LGTM

PR #122, opened 2021-01-05
Ocaml 4.8

Enkelmann:
This PR updates the Ocaml source files to support Ocaml 4.08 and BAP 2.2. The version number was also changed to 0.4, since this is probably the last PR before officially releasing version 0.4.


mellowCS (approved):
LGTM!

PR #123, opened 2021-01-06
Backward analysis trait

mellowCS:
This PR contains the backward interprocedural fixpoint analysis including the test infrastructure. Furthermore a generic interprocedural fixpoint file was added to reduce code duplication.


cwe_checker_rs/src/analysis/backward_interprocedural_fixpoint/tests.rs
```diff
@@ -0,0 +1,245 @@
...
+    assert_eq!(
+        *computation
+            .get_node_value(NodeIndex::new(0))
+            .unwrap()
+            .unwrap_value(),
+        0 as u64
+    );
+    assert_eq!(
+        *computation
+            .get_node_value(NodeIndex::new(1))
```
Enkelmann, line 170: Hmm... This could fail if some unrelated change in `graph.rs` also changes the order in which nodes are added to the graph. Why not identify them by the block TIDs instead?

cwe_checker_rs/src/analysis/interprocedural_fixpoint_generic.rs
```diff
@@ -0,0 +1,36 @@
+use crate::prelude::*;
+
+/// NodeValue that can either be a simple value or a
+/// combination of two values representing a fall through
+/// and interprocedural flow value
+#[derive(PartialEq, Eq, Serialize, Deserialize)]
+pub enum NodeValue<T: PartialEq + Eq> {
+    Value(T),
+    // TODO: more doc strings
```
Enkelmann, line 9: TODO. ;-)

Enkelmann (changes requested):

cwe_checker_rs/src/analysis/backward_interprocedural_fixpoint/mod.rs
```diff
@@ -1,11 +1,24 @@
...
 use super::fixpoint::Context as GeneralFPContext;
 use super::graph::*;
 use super::interprocedural_fixpoint_generic::*;
 use crate::intermediate_representation::*;
-use fnv::FnvHashMap;
-use petgraph::graph::{EdgeIndex, NodeIndex};
+use petgraph::graph::EdgeIndex;
 use std::marker::PhantomData;
 
+pub mod mock_context;
```
Enkelmann: The mock context is only used for unit tests, so we should only compile it when executing tests by adding `#[cfg(test)]`.

cwe_checker_rs/src/analysis/backward_interprocedural_fixpoint/mod.rs
```diff
@@ -43,8 +56,8 @@ pub trait Context<'a> {
     /// Transition function for in-program calls.
     fn update_callsite(
         &self,
-        value_after_call: Option<&Self::Value>,
-        fallthrough_value: Option<&Self::Value>,
+        target_value: Option<&Self::Value>,
+        return_value: Option<&Self::Value>,
```
Enkelmann, line 60: Hmm, I am not sure whether these new names are better. But since I don't know better names either, we probably should simply add a clarifying comment to the doc comment of `update_callsite`.

Enkelmann (changes requested):

cwe_checker_rs/src/analysis/backward_interprocedural_fixpoint/mod.rs
```diff
@@ -54,6 +52,8 @@ pub trait Context<'a> {
     ) -> Option<Self::Value>;
 
     /// Transition function for in-program calls.
+    /// The target value is comming in via the call edge from the called subroutine and
+    /// the return_value is comming in via the call stub edge from the return node of the subroutine
```
Enkelmann, line 56: it is coming from the returned-to site of the caller, right? Not from the returned-from site of the callee like in the forward analysis. This is why I am being so insistent on being precise here...

cwe_checker_rs/src/analysis/backward_interprocedural_fixpoint/tests.rs
```diff
@@ -0,0 +1,299 @@
...
+            name: String::from("RSP"),
+            size: ByteSize::new(8),
+            is_temp: false,
+        },
+        calling_conventions: Vec::new(),
+    };
+
+    let mock_con = Context::new(&project);
+    let mut computation = create_computation(mock_con.clone(), None);
+    computation.set_node_value(NodeIndex::new(0), NodeValue::Value(0));
```
Enkelmann, line 151: You missed this one, which is still dependent on the order of node insertions in `graph.rs`.

cwe_checker_rs/src/analysis/backward_interprocedural_fixpoint/mock_context.rs
```diff
@@ -0,0 +1,126 @@
...
+pub struct Context<'a> {
+    pub graph: Graph<'a>,
+    pub tid_to_node_index: HashMap<(Tid, Tid, StartEnd), NodeIndex>,
+}
+
+impl<'a> Context<'a> {
+    pub fn new(project: &'a Project) -> Self {
+        let mut graph = crate::analysis::graph::get_program_cfg(&project.program, HashSet::new());
+        let mut tid_to_node_index: HashMap<(Tid, Tid, StartEnd), NodeIndex> = HashMap::new();
+        graph.reverse();
```
Enkelmann, line 24: I almost missed this one. My hope was that the user does not have to explicitly reverse the control flow graph when defining a context for a backwards analysis, so that forward and backward analyses could be performed on the same control flow graph. We should check together whether we can achieve this somehow.

Enkelmann (changes requested):
By the way, I checked your unit test by hand and everything is correct, so the backwards analysis works as intended. :-)

cwe_checker_rs/src/analysis/backward_interprocedural_fixpoint/mock_context.rs
mellowCS: We could simply add a parameter 'backward: bool' to the get_program_cfg() and reverse the graph in that function. However, the user would have to add that parameter as well.

Enkelmann (approved):

PR #124, opened 2021-01-08
move the new_top() function to the HasByteSize trait.

Enkelmann:
This PR moves the `new_top()` function from the `RegisterDomain` trait to the `HasByteSize` trait. This way we can relax the conditions for the values inside a `MemRegion` from implementing `RegisterDomain` to just implementing `HasTop` and `HasByteSize`.


mellowCS (approved):
LGTM!

PR #126, opened 2021-01-12
Cargo test action

Enkelmann:
This PR adds Github actions for running unit tests and acceptance tests.

The acceptance tests are running way too slow right now, thus they only run on pull requests right now. Part of the reason is that we cannot run them in parallel without triggering random test failures. Execution of Ghidra for very small input files is also way too slow at the moment, but this is a problem for another PR.

Edit: The PR now also contains a bugfix for the CWE 243 check uncovered by the acceptance tests.


mellowCS (approved):
LGTM!

PR #128, opened 2021-01-14
Ghidra switch

Enkelmann:
This PR switches the default backend for the *cwe_checker* from the old BAP backend to the new Ghidra backend.


README.md
```diff
@@ -45,82 +64,81 @@ If you plan to develop *cwe_checker*, it is recommended to build it using the pr
...
-The Ghidra backend is still in early development, thus some checks are not yet available for it. To try it out, the following dependencies must be fulfilled:
--   [Rust](https://www.rust-lang.org) >= 1.44.1
--   [Ghidra](https://ghidra-sre.org/) >= 9.2. If you want to use an earlier version of Ghidra, you need to manually add the Java library `gson` to Ghidra: Download it from https://search.maven.org/artifact/com.google.code.gson/gson/2.8.6/jar and move it to the Ghidra plugin folder located at `$HOME/.ghidra/.ghidra_9.X.X_PUBLIC/plugins` (with the version number depending on your version of Ghidra).
-
-Run `make all GHIDRA_PATH=path/to/ghidra_folder` (with the correct path to the local Ghidra installation inserted) to compile and install the *cwe_checker*.
+Just run `make with_bap_backend` to compile and register the plugin with BAP.
 
 ## Usage ##
 
-The *cwe_checker* takes as input a binary file, runs several [checks](#checks) based on static analysis on the binary and then outputs a list of CWE warnings that have been found during the analysis.
```
0xricksanchez, line 62: The *cwe_checker* takes a binary as input, ...?

README.md
```diff
@@ -45,82 +64,81 @@ If you plan to develop *cwe_checker*, it is recommended to build it using the pr
...
 -   [CWE-782](https://cwe.mitre.org/data/definitions/782.html): Exposed IOCTL with Insufficient Access Control
 
-Please note that some of the above analyses only are partially implemented at the moment. Furthermore, false positives are to be expected due to shortcuts and the nature of static analysis as well as over-approximation. For more information about the individual checks you can look at the [online documentation](https://fkie-cad.github.io/cwe_checker/doc/html/cwe_checker/index.html).
-
-**NOTE:** The new memory checks for CWEs 415 and 416 are still very experimental and are disabled on a standard run. You can try them out using the `-partial=Memory` command line flag.
-
-**NOTE:** We recently decided to deprecate the support for the old emulation based checks for CWEs 415, 416 and 787. In addition to trying out the new memory checks, users of these checks should also take a look at the [BAP toolkit](https://github.com/BinaryAnalysisPlatform/bap-toolkit), which provides better-maintained (and still emulation based) versions of these checks.
+Please note that some of the above analyses only are partially implemented at the moment.
```
0xricksanchez, line 115: Should we note which ones are, and if yes detail which functionality is missing?
On the same note: Highlight plans about which further checks (if any) are on the Roadmap next?

README.md
```diff
@@ -45,82 +64,81 @@ If you plan to develop *cwe_checker*, it is recommended to build it using the pr
...
-## How to extend cwe_checker? ##
-You can find some information about how to write your own check [here](https://fkie-cad.github.io/cwe_checker/doc/html/cwe_checker/index.html#HackingHowto)
-
-If you plan to open a PR, please utilize [precommit](https://pre-commit.com) in your development environment to catch many issues before the code review.
-
-### Contribute ###
-Contributions are always welcome. Just fork it and open a pull request!
+*cwe_checker* comes with a script for Ghidra,
+which parses the output of the *cwe_checker* and annotates the found CWEs in the disassembler for easier manual analysis.
+The script is located at `ghidra_plugin/cwe_checker_ghidra_plugin.py`, usage instructions are contained in the file.
```
0xricksanchez, line 122: I think a Demo/Screenshot would be cool after all. Breaks up the Wall of text.
Maybe a short screencast instead of a Screenshot to see the Tool in action (incl. its Ghidra integration!

README.md
```diff
@@ -8,17 +8,28 @@
...
+These bug classes are formally known as [Common Weakness Enumerations](https://cwe.mitre.org/) (CWEs).
+Its main goal is to aid analysts to quickly find vulnerable code paths.
 
-Its main focus are ELF binaries that are commonly found on Linux and Unix operating systems. *cwe_checker* is built on top of [BAP](https://github.com/BinaryAnalysisPlatform/bap) (Binary Analysis Platform). By using BAP, we are not restricted to one low level instruction set architectures like Intel x86. BAP lifts several of them to one common intermediate representation (IR). cwe_checker implements its analyses on this IR. At time of writing, BAP 2.2 supports Intel x86/x64, ARM, MIPS, and PPC amongst others. Hence, this makes *cwe_checker* a valuable tool for firmware analysis.
+Its main focus are ELF binaries that are commonly found on Linux and Unix operating systems.
+The *cwe_checker* uses [Ghidra](https://ghidra-sre.org/) to disassemble binaries into one common intermediate representation
+and implements its own analyses on this IR.
+Hence, the analyses can be run for all CPU architectures that Ghidra can disassemble,
```
0xricksanchez, line 24: ~~for~~ on

0xricksanchez (commented):
Minor nitpicking at best. Rest lgtm.

ivg:
Are you guys no longer dependent on BAP at all, or you can still use BAP as an alternative backend?

Enkelmann:
@ivg One can still use the BAP backend (after merging of this PR with `make with_bap_backend`).

However, we do not have the manpower to actively develop for both backends and right now the Ghidra backend is simply better for our needs. So the current plan is to phase out support for BAP completely in the near future.

README.md
Enkelmann: As long as we do not hit version 1.0 I would like to keep it somewhat vague to emphasize that one cannot blindly trust the results yet. Check-specific notes are contained in the documentation one can build with `make documentation`, but I still need to make some minor improvements to it before I link to it here. This will be done in another PR.

We do not have a public Roadmap right now.

README.md
Enkelmann: You are right. I will look into it. Where would you place it? Right at the beginning in the "What is cwe_checker"-section?

README.md
Enkelmann: fixed.

README.md
Enkelmann: fixed

README.md
0xricksanchez: I think I'd put it right before the Installation section. So the interested user can read up on what the tool does up to that point and then get a clear image of what it may look like when using the tool.

Enkelmann:
@0xricksanchez I added two pictures to somewhat lighten up the wall of text. What do you think of them?
I had to scratch my idea with creating an asciinema for the Readme, since it would create unreasonable expectations for the runtime of the cwe_checker.

0xricksanchez (approved):
Liking that README a lot more with the added pictures. I think the possible workflows (CLI vs Ghidra integration) is clear enough with those and an Asciinema is not needed.

mellowCS (approved):
LGTM!

PR #129, opened 2021-01-19
Add address_base_offset field to Program struct

Enkelmann:
This PR adds an `address_base_offset` field to the program struct. This is the difference between addresses as determined by the binary headers and addresses as determined by Ghidra.

This will be used to enable access to the correct addresses in global memory for analyses (in a future PR).


mellowCS (approved):
LGTM!

PR #130, opened 2021-01-21
CWE 78

mellowCS:
This PR adds the CWE 78 check for OS Command Injections


cwe_checker_rs/src/checkers/cwe_78/context/mod.rs
```diff
@@ -0,0 +1,563 @@
...
+                    );
+                }
+            }
+        }
+
+        new_state
+    }
+
+    /// Checks whether the current def term is the last def term
+    /// of its corresponding block and if so, returns the block term and correspondoing
```
Enkelmann, line 300: this doc comment seems to be unfinished

cwe_checker_rs/src/checkers/cwe_78/context/mod.rs
```diff
@@ -0,0 +1,563 @@
...
+            {
+                let mut pi_def_map: HashMap<Tid, PointerInferenceState> = HashMap::new();
+
+                let pi_context = self.pointer_inference_results.get_context();
+                let mut new_pi_state = Some(pi_value.unwrap_value().clone());
+
+                for def in block_node.get_block().term.defs.iter() {
+                    // Add the pi state to the map that is available after the def was executed
+                    // If no state is available after the update_def() call, none is added
+                    new_pi_state = pi_context.update_def(&new_pi_state.unwrap(), def);
```
Enkelmann, line 336: If update_def returns `None` before the last def of the block, then the `new_pi_state.unwrap()` would panic during the next iteration. We should `break` out of the loop before that happens.

cwe_checker_rs/src/checkers/cwe_78/context/mod.rs
```diff
@@ -0,0 +1,563 @@
...
+            Expression::Var(ram_target) => ram_target,
+            Expression::Subpiece { arg, .. } => {
+                let argument: &Expression = arg;
+                match argument {
+                    Expression::Var(ram_target) => ram_target,
+                    _ => panic!("Unexpected target format for Store instruction!"),
+                }
+            }
+            _ => panic!("Unexpected target format for Store instruction!"),
+        };
```
Enkelmann, line 393: I don't think we can rely on address expressions having this particular form. E.g. we may add a pass to remove unnecessary temporary registers, which could lead this code to panic in the process. So we should either try to handle generic address expressions, or at least prevent this code from panicking if the address expression is not as expected. Also: What about constant addresses, i.e. addresses of global variables?

cwe_checker_rs/src/checkers/cwe_78/context/mod.rs
```diff
@@ -0,0 +1,563 @@
...
+            | Def::Load {
+                var,
+                address: input,
+            } => new_state = self.handle_assign_and_load(new_state, def, var, input),
+            Def::Store { address, value } => {
+                new_state = self.handle_store(new_state, def, address, value)
+            }
+        }
+
+        Some(new_state)
```
Enkelmann, line 450: I think we should also check whether the `def` is the first node of the block. In that case we could remove the `pi_def_map` from the `new_state` and thus preventing unnecessary information to be stored in the computation struct.

cwe_checker_rs/src/checkers/cwe_78/context/mod.rs
```diff
@@ -0,0 +1,563 @@
...
+        &self,
+        state_after_jump: &State,
+        _jump: &Term<Jmp>,
+        _untaken_conditional: Option<&Term<Jmp>>,
+        _jumpsite: &Term<Blk>,
+    ) -> Option<State> {
+        Some(state_after_jump.clone())
+    }
+
+    fn update_callsite(
```
Enkelmann, line 465: missing doc comment

cwe_checker_rs/src/checkers/cwe_78/context/mod.rs
```diff
@@ -0,0 +1,563 @@
...
+            let mut new_state = return_.clone();
+            if let Some(calling_conv) = self.project.get_standard_calling_convention() {
+                new_state.remove_non_callee_saved_taint(calling_conv);
+            }
+
+            return Some(new_state);
+        }
+
+        // If there no return state, there will also be no target state and therefore None is returned
+        None
```
Enkelmann, line 503: What if the taint source/system call was inside the callee? Then there would be a target_state without a return_state.

cwe_checker_rs/src/checkers/cwe_78/context/mod.rs
```diff
@@ -0,0 +1,563 @@
...
+        let mut new_state = combined_state.clone();
+        if let Some(calling_conv) = self.project.get_standard_calling_convention() {
+            let return_registers: HashSet<String> =
+                calling_conv.return_register.iter().cloned().collect();
+            new_state.remove_all_except_return_register_taints(return_registers);
+        }
+        Some(new_state)
+    }
+
+    /// Check whether the term is a direkt or indirekt call. If it is direct, check for the
```
Enkelmann, line 523: typo: `direkt` -> `direct` (twice)

cwe_checker_rs/src/checkers/cwe_78/context/mod.rs
```diff
@@ -0,0 +1,563 @@
...
+    /// to distinguish the nodes for blocks contained in more than one function.
+    jmp_to_blk_end_node_map: Arc<HashMap<(Tid, Tid), NodeIndex>>,
+    /// The call whose parameter values are the sources for taint for the analysis.
+    pub taint_source: Option<&'a Term<Jmp>>,
+    /// The name of the function, whose parameter values are the taint sources.
+    pub taint_source_name: Option<String>,
+    /// The current subfunction.
+    ///Since the analysis is intraprocedural,
+    ///all nodes with state during the fixpoint algorithm should belong to this function.
+    current_sub: Option<&'a Term<Sub>>,
```
Enkelmann, line 52: Is the analysis really intraprocedural? You do propagate taint across edges not belonging to your subfunction if I read the context trait implementation correctly.

Enkelmann (changes requested):
Whew, that is a lot of code!
It should be possible to use the same taint domain as the CWE 476 check, thus preventing some code duplication.
I did not review the unit tests yet. I'll do that for the re-review.

cwe_checker_rs/src/checkers/cwe_78/context/mod.rs
mellowCS: Oops. Must have misread that when I copied the comment.

cwe_checker_rs/src/checkers/cwe_78/state/mod.rs
```diff
@@ -0,0 +1,417 @@
...
+#[derive(Serialize, Deserialize, Debug, Eq, Clone)]
+pub struct State {
+    /// The set of currently tainted registers.
+    register_taint: HashMap<Variable, Taint>,
+    /// The Taint contained in memory objects
+    memory_taint: HashMap<AbstractIdentifier, MemRegion<Taint>>,
+    /// The set of addresses in the binary where strings constants reside
+    string_constants: Vec<Bitvector>,
+    /// A map from Def Tids to their corresponding pointer inference state
+    /// The pointer inferenece states are calculated in a forward manner
```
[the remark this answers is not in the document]
Enkelmann: You forgot to fix the `inferenece => inference` typo.

cwe_checker_rs/src/checkers/cwe_78/state/mod.rs
```diff
@@ -0,0 +1,417 @@
...
+                    mem_region.insert_at_byte_index(*taint, *index);
+                    // Unsound in theory for partially intersecting taints. Should not matter in practice.
+                }
+            } else {
+                memory_taint.insert(tid.clone(), other_mem_region.clone());
+            }
+        }
+
+        let mut constants = self.string_constants.clone();
+        constants.extend(other.string_constants.clone());
```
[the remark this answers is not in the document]
Enkelmann: Ok, deduplication is one solution. But why not make `string_constants` a `HashSet` or a `BTreeSet` in the first place?

cwe_checker_rs/src/intermediate_representation/expression/builder.rs
```diff
@@ -1,8 +1,47 @@
-use super::{BinOpType, Expression};
+use apint::ApInt;
+
+use super::{BinOpType, CastOpType, Expression, Variable};
 use crate::prelude::*;
 
 /// ## Helper functions for building expressions
 impl Expression {
+    // Shortcut for creating a constant expression from an i64 value
```
Enkelmann, line 8: Doc comments start with `///`, not with `//`. This also applies to the other functions in this module.

cwe_checker_rs/src/intermediate_representation/expression/builder.rs
```diff
@@ -1,8 +1,47 @@
...
+    }
+
+    // Shortcut for creating a cast expression
+    pub fn cast(self, op: CastOpType) -> Expression {
+        Expression::Cast {
+            op,
+            size: ByteSize::new(8),
+            arg: Box::new(self),
+        }
+    }
```
Enkelmann, line 34: We should either add `size` as a parameter to both functions or make them `#[cfg(test)]`, since the implicit size assumption of 8 bytes would introduce very hard to spot errors if used in non-unit-test-code.

Enkelmann (changes requested):
I will split my review into several parts this time. Here is the first part. :-)

cwe_checker_rs/src/intermediate_representation/term/builder.rs
```diff
@@ -0,0 +1,61 @@
...
+                },
+            };
+        }
+        Term {
+            tid: Tid::new(tid),
+            term: Jmp::Call {
+                target: Tid::new(target_tid),
+                return_: None,
+            },
+        }
```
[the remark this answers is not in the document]
mellowCS: Does this work if return_tid is None?

Enkelmann (changes requested):
Part 2 of the review.

cwe_checker_rs/src/intermediate_representation/term/builder.rs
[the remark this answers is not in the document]
Enkelmann: Yes, since the `map` method of `Option<T>` maps `None` to `None`.

Enkelmann (changes requested):
... and here the third and final part of the review. :-)

The CWE messages should contain more information (like where the source for the potential user input is), but I think we should postpone that to the next PR for this check, since we will have more useful information to present to the user then compared to what we have now.

cwe_checker_rs/src/checkers/cwe_78/context/mod.rs
```diff
@@ -0,0 +1,607 @@
...
+            if new_state.check_return_registers_for_taint(return_registers) {
+                new_state
+                    .remove_non_callee_saved_taint(symbol.get_calling_convention(self.project));
+                // TODO: Parameter detection since targets of input parameters are the return locations
+                // Taint memory for string inputs
+                if symbol.name == "scanf" {
+                    self.generate_cwe_warning(
+                        &new_state.get_current_sub().as_ref().unwrap().term.name,
+                    );
+                }
```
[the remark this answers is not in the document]
mellowCS: It was meant for the next PR, so I won't forget.

cwe_checker_rs/src/checkers/cwe_78/state/mod.rs
```diff
@@ -0,0 +1,465 @@
...
+            | Expression::Cast { arg, .. }
+            | Expression::Subpiece { arg, .. } => {
+                self.taint_def_input_register(arg, stack_pointer_register, def_tid)
+            }
+            _ => (),
+        }
+    }
+
+    /// Either taints the input register or a memory position if it is the stack pointer register
+    pub fn taint_variable_input(
```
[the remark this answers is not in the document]
mellowCS: This is actually complete. It means that we taint the memory if the input register is the stack pointer. I could rephrase if you want.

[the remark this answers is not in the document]
Enkelmann: Not necessary, I was just unable to read it properly. ;-) But you could add a dot at the end of the sentence.

cwe_checker_rs/src/checkers/cwe_78/context/mod.rs
```diff
@@ -121,8 +124,8 @@ impl<'a> Context<'a> {
         let source = self.taint_source.unwrap();
         let name = self.taint_source_name.clone().unwrap();
         let description: String = format!(
-            "(Potential OS Command Injection) {} ({}) -> {}",
-            sub_name, source.tid.address, name
+            "(Input for call {} is not properly sanitized) {} ({}) -> {}",
+            name, sub_name, source.tid.address, name
```
Enkelmann, line 128: I didn't mean that the `(Potential OS Command Injection)` should be removed. More like
```
(Potential OS Command Injection) Input for call to {} is not properly sanitized in function {} ({})
```
i.e. combine the two. :-) This would also be the same style as the other CWE warnings.

Enkelmann (changes requested):
There is one issue you forgot from the previous review. The rest is almost ready to be merged! :-)

Enkelmann (approved):
Now everything looks fine! :-)
I will merge it tomorrow if the CI also approves.

PR #131, opened 2021-01-25
Add runtime memory image representation

Enkelmann:
This PR adds a struct representing the runtime memory image of the binary. 

@mellowCS Since we will use this struct to access global (read-only) variables, please double-check whether the interface covers your use-case. If not, I can extend the interface.


mellowCS (approved):
LGTM!

PR #132, opened 2021-01-26
Cross compiling

Enkelmann:
This PR updates the cross compiler toolchain to use Python3.

The old toolchain probably broke down since Python2 is now officially deprecated.


mellowCS (approved):
LGTM!

PR #133, opened 2021-01-27
Pi global mem

Enkelmann:
This PR adds handling of stores and loads to global memory for the pointer inference analysis.

This slightly increases exactness of the analysis and removes a lot of unnecessary debug messages, which were generated for global memory accesses.


mellowCS (approved):
LGTM!

PR #134, opened 2021-01-27
change base image of dockerfile

Enkelmann:
This PR changes the base image for the docker image to one where Ghidra is already installed. This makes the image a little bit smaller and may also save some build time for the docker image.


mellowCS (approved):
LGTM!

PR #135, opened 2021-01-28
Bugfix Stack Register Computation

Enkelmann:
This PR fixes a bug, where the stack register was overwritten on return instructions with its value before the corresponding call. This produced incorrect stack pointer values for x86 and x64 architectures.


mellowCS (approved):
LGTM!

PR #138, opened 2021-02-02
Trivial expressions

Enkelmann:
Substitute more examples of trivial expressions with their results. Also enable more acceptance tests.

It turns out that most of our MIPS-based analyses only failed because the analysis did not know that `a or 0` is always `a`. This is now fixed.

### DIFF
--- a/test/src/lib.rs
+++ b/test/src/lib.rs
@@ -174,13 +174,6 @@ mod tests {
         let mut error_log = Vec::new();
         let mut tests = all_test_cases("cwe_190", "CWE190");
 
-        // Ghidra does not recognize all extern function calls in the disassembly step for MIPS.
-        // Needs own control flow graph analysis to be fixed.
-        mark_skipped(&mut tests, "mips64", "clang");
-        mark_skipped(&mut tests, "mips64el", "clang");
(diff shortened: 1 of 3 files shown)


test/src/lib.rs
```diff
@@ -395,21 +367,12 @@ mod tests {
         mark_skipped(&mut tests, "arm", "clang");
         mark_skipped(&mut tests, "mips", "clang");
         mark_skipped(&mut tests, "mipsel", "clang");
-
-        // Ghidra does not recognize all extern function calls in the disassembly step for MIPS.
-        // Needs own control flow graph analysis to be fixed.
         mark_skipped(&mut tests, "mips64", "clang");
```
mellowCS, line 370: Did you find out why the clang compiled files still fail?

mellowCS (approved):

test/src/lib.rs
Enkelmann: For the same reason as for the ARM-based clang test cases: The constant that the check looks for is not defined in the basic block of the corresponding call. Since the check only looks in the basic block before the call, it cannot find the constant.

PR #140, opened 2021-02-04
better error messages on failed Ghidra execution

Enkelmann:
Better error messages if the execution of Ghidra fails. Also prevents printing a stack backtrace in this case because it is irrelevant.


mellowCS (approved):
LGTM!

Issue #141, opened 2021-02-04
Issues running cwe_checker


alexdetrano:
Hi there, 
I am very interested in this tool but am having some problems running it. I started off using the docker method and I get the following error:

```
/tmp > docker run --rm -v /tmp/socat:/input fkiecad/cwe_checker /input
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error("unknown variant `ROUND`, expected one of `COPY`, `LOAD`, `STORE`, `PIECE`, `SUBPIECE`, `POPCOUNT`, `INT_EQUAL`, `INT_NOTEQUAL`, `INT_LESS`, `INT_SLESS`, `INT_LESSEQUAL`, `INT_SLESSEQUAL`, `INT_ADD`, `INT_SUB`, `INT_CARRY`, `INT_SCARRY`, `INT_SBORROW`, `INT_XOR`, `INT_AND`, `INT_OR`, `INT_LEFT`, `INT_RIGHT`, `INT_SRIGHT`, `INT_MULT`, `INT_DIV`, `INT_REM`, `INT_SDIV`, `INT_SREM`, `BOOL_XOR`, `BOOL_AND`, `BOOL_OR`, `FLOAT_EQUAL`, `FLOAT_NOTEQUAL`, `FLOAT_LESS`, `FLOAT_LESSEQUAL`, `FLOAT_ADD`, `FLOAT_SUB`, `FLOAT_MULT`, `FLOAT_DIV`, `INT_NEGATE`, `INT_2COMP`, `BOOL_NEGATE`, `FLOAT_NEG`, `FLOAT_ABS`, `FLOAT_SQRT`, `FLOAT_CEIL`, `FLOAT_FLOOR`, `FLOAT_ROUND`, `FLOAT_NAN`, `INT_ZEXT`, `INT_SEXT`, `INT2FLOAT`, `FLOAT2FLOAT`, `TRUNC`", line: 45658, column: 45)', caller/src/main.rs:277:64
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
The file type is a 32-bit Linux ELF for ARM:
```
/tmp > file /tmp/socat
/tmp/socat: ELF 32-bit LSB executable, ARM, EABI5 version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-armhf.so.3, for GNU/Linux 3.2.0, BuildID[sha1]=5cfa8023dbbef69e608816f9967e981c94910f20, stripped
[...]

PR #143, opened 2021-02-08
Correctly deserialize FLOAT_ROUND p-code operations

Enkelmann:
Apparently, Ghidra uses ROUND as a mnemonic for the FLOAT_ROUND p-code operation. At least according to the p-code op code it represents the FLOAT_ROUND p-code operation. This PR fixes the deserialization of it and should be a fix for issue #141.


mellowCS (approved):
LGTM!

PR #145, opened 2021-02-10
Ghidra pipe

mellowCS:
This PR enables running Ghidra in a separate thread while keeping its output in memory by piping its output via a FIFO  to Rust for further processing.


caller/src/main.rs
```diff
@@ -241,51 +243,80 @@ fn get_project_from_ghidra(file_path: &Path, binary: &[u8], quiet_flag: bool) ->
...
+            Ok(_) => contents,
+            Err(err) => panic!("Failed to read contents from fifo pipe {}", err),
+        }
+    } else {
+        panic!("Could not open fifo pipe after contents have been written by Ghidra.");
+    };
+
     let mut project_pcode: cwe_checker_rs::pcode::Project =
-        serde_json::from_reader(std::io::BufReader::new(file)).unwrap();
+        serde_json::from_str(json_string.as_str()).unwrap();
```
Enkelmann, line 319: Did the `serde_json::from_reader` not work with the pipe as input?

caller/src/main.rs
```diff
@@ -301,7 +332,10 @@ fn get_project_from_ghidra(file_path: &Path, binary: &[u8], quiet_flag: bool) ->
             project
         }
     };
-    // delete the temporary file again.
-    std::fs::remove_file(output_path).unwrap();
```
Enkelmann, line 305: We should also remove the pipe if we don't need it anymore.

Enkelmann (changes requested):
The PR already leads to a significant speed up of the acceptance test suite, so it is definitely worth it! :-)

Enkelmann (approved):
Looks good. I will do an additional stress test before merging, just to be sure.

Enkelmann:
The stress test showed that I missed a debug print during my review. ;-) Other than that it worked without any problems. I had to disable the CWE 78 check for standard runs though, because the check used up huge amounts of RAM and CPU time on some binaries. We should investigate that.

PR #151, opened 2021-02-19
parse indirect jump targets from Ghidra

Enkelmann:
Parse indirect jump targets from the Ghidra backend and add them to the control flow graph. This PR only handles intraprocedural indirect jumps but not indirect calls!


mellowCS (approved):
LGTM!

PR #163, opened 2021-04-13
Fix missing field tid bug

mellowCS:
This PR fixes the bug where no TID was given to extern symbols which have a reference to themselves.


Enkelmann (approved):
LGTM

PR #164, opened 2021-04-14
Memory string reader

mellowCS:
This PR contains a small extension to the Runtime Memory Image. It allows to read bytes until a null terminator.


src/cwe_checker_lib/src/utils/binary.rs
```diff
@@ -167,6 +167,30 @@ impl RuntimeMemoryImage {
...
+        for segment in self.memory_segments.iter() {
+            if address >= segment.base_address
+                && address <= segment.base_address + segment.bytes.len() as u64
+            {
+                let index = (address - segment.base_address) as usize;
+                let mut bytes: Vec<u8> = Vec::new();
+                for byte in segment.bytes[index..].iter() {
+                    if *byte == 0 {
+                        if self.is_little_endian {
+                            bytes = bytes.into_iter().rev().collect();
```
Enkelmann, line 182: This is wrong: The representation of strings is **not** affected by the endianess of the architecture.

Enkelmann (changes requested):
Instead of allocating `Vec<u8>` for the resulting string, I would use `std::ffi::CStr::from_bytes_with_nul`, which should do exactly what you want and returns a borrowed variant of the string instead of an owned representation.

src/cwe_checker_lib/src/utils/binary.rs
```diff
@@ -167,6 +167,22 @@ impl RuntimeMemoryImage {
         Err(anyhow!("Address is not a valid global memory address."))
     }
 
+    /// Read the contents of memory from a given address onwards until a null byte is reached.
```
Enkelmann, line 170: The doc comment is slightly outdated, since the current version additionally checks whether the byte sequence is a valid utf-8 string.

Enkelmann (changes requested):
Code looks good. Comments need a little update before I merge it.

Enkelmann (approved):

PR #165, opened 2021-04-15
Bugfixes

### LINKED ISSUE #161: Panic due to "missing field `tid`"
I tried to run the docker image on the [botan](https://github.com/randombit/botan) cli executable. Unfortunately, after running for a while it panics with
```
$ docker run --rm -v $(pwd)/botan:/input fkiecad/cwe_checker /input
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error("missing field `tid`", line: 80734887, column: 9)', src/caller/src/main.rs:300:64
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
Checking a very simple hello-world example works fine, though.

I'm on Manjaro Linux with kernel `5.10.23-1`.

Enkelmann:
Fixes two bugs:
- When Ghidra reports a function without a basic block at that address, we now report it as a log message and treat the function as a dead end instead of panicking.
- The byte size of the `FloatNaN` unary operation was sometimes incorrectly set to the input byte size. Now it is correctly set to 1.

Closes #161.

### DIFF
--- a/src/caller/src/main.rs
+++ b/src/caller/src/main.rs
@@ -113,9 +113,9 @@ fn run_with_ghidra(args: CmdlineArgs) {
             binary_file_path.display()
         )
     });
-    let mut project = get_project_from_ghidra(&binary_file_path, &binary[..], args.quiet);
+    let (mut project, mut all_logs) = get_project_from_ghidra(&binary_file_path, &binary[..]);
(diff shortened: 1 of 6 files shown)


mellowCS (approved):
LGTM

PR #166, opened 2021-04-20
Widening delay

Enkelmann:
Add a widening delay to the widening strategy to prevent unnecessary widenings. Also fixes some bugs related to the generation of bound hints.


mellowCS (approved):
LGTM

PR #167, opened 2021-04-20
Callee saved return targets

mellowCS:
This PR simplifies the taint of function arguments and improves return value detection by temporarily adding pointer targets of callee saved registers to the memory taints before a string function call.


Enkelmann (approved):
LGTM

PR #168, opened 2021-04-21
Implicit load in indirect calls

Enkelmann:
Implicit load instructions in the target expression of indirect jumps and calls are now correctly converted to explicit load instructions. We already did that for implicit loads in `Def` instructions, but apparently forgot to implement it for jumps.


mellowCS (approved):
LGTM

PR #170, opened 2021-04-27
fix address size of implicit stores

Enkelmann:
Fixes a bug, where the address size of store instructions was in some cases incorrectly truncated to the byte size of the value to be stored.


mellowCS (approved):
LGTM

Issue #169, opened 2021-04-21
busybox binary still getting error.


anzosasuke:
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/cwe_checker_lib/src/analysis/graph.rs:312:85
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

Enkelmann:
Can you provide the affected binary? None of my busybox samples trigger this bug. And since it is very likely that the root cause of the bug is not where the crash occurs, I probably need a sample to find the root cause.

anzosasuke:
here it is.
[busybox.zip](https://github.com/fkie-cad/cwe_checker/files/6359023/busybox.zip)

Enkelmann:
Thanks! I can confirm the issue, we are working on a fix.

Enkelmann:
Short update: The original bug is fixed thanks to PR #171. But we encountered a second bug afterwards, which we are currently fixing.

Enkelmann:
The second bug is now also fixed thanks to PR #172.

Apparently we still have problems handling the MIPS delay slots in some cases (which results in a lot of log messages for the busybox binary). At the moment we are debating whether we want to make a larger rewrite of the affected code to fix this and similar issues, so I do not know yet when exactly we will tackle this problem. But this is definitely not part of the original issue here anymore. ;-)

PR #171, opened 2021-04-27
- Fixed issue #169

mellowCS:
- Fixed issue #169 by checking whether the Call has a fallThrough address. Now no Return value is given to the Call.
- Fixed bug where call return pairs were not properly handled when branch instructions were added to the pcode instructions.
  * Adding the branch instruction caused the running pcode index to increase but not the number of pcode instructions which
    needs to be correct for the CallReturn Handler to be triggered.
- Removed dead parameter from jump processing function.


Enkelmann (approved):
LGTM

PR #172, opened 2021-04-28
Fix bug in sub register handling

Enkelmann:
- The position of sub registers in the base register was sometimes incorrect for 32bit architectures.
- A special case for sub register handling was missing.


mellowCS (approved):
LGTM

PR #173, opened 2021-05-03
- Added data type properties to the project struct that contain the s…

mellowCS:
This PR includes the following: 
    - Added data type properties to the project struct that contain the sizes of data types in bytes
    - Added a hasVarArgs flag to external symbols to indicate whether an external symbol has a variable amount of parameters
    - Made the appropriate changes on the Rust side.

### DIFF
--- a/src/cwe_checker_lib/src/pcode/expressions.rs
+++ b/src/cwe_checker_lib/src/pcode/expressions.rs
@@ -371,6 +371,29 @@ pub struct RegisterProperties {
     pub size: ByteSize,
 }
 
+/// Properties of data types such as size.
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Hash, Clone)]
+pub struct DatatypeProperties {
+    /// Holds the size of the char type
+    pub char_size: ByteSize,
+    /// Holds the size of the double type
+    pub double_size: ByteSize,
+    /// Holds the size of the float type
+    pub float_size: ByteSize,
+    /// Holds the size of the integer type
+    pub integer_size: ByteSize,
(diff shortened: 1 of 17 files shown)


src/cwe_checker_lib/src/pcode/expressions.rs
```diff
@@ -371,6 +371,29 @@ pub struct RegisterProperties {
...
+    /// Holds the size of the long long type
+    pub long_long_size: ByteSize,
+    /// Holds the size of the long type
+    pub long_size: ByteSize,
+    /// Holds the size of the pointer type
+    pub pointer_size: ByteSize,
+    /// Holds the size of the short type
+    pub short_size: ByteSize,
+}
+
```
Enkelmann, line 396: This type definition should be in the `intermediate_representation` module, not the `pcode` module. Since the intermediate representation should abstract away the pcode representation, every type that is shared by both representations (e.g. `Tid`) should be defined in the `intermediate_representation` module. This way one never needs to care about anything in the `pcode` module when writing analyses.

src/cwe_checker_lib/src/pcode/expressions.rs
```diff
@@ -371,6 +371,29 @@ pub struct RegisterProperties {
     pub size: ByteSize,
 }
 
+/// Properties of data types such as size.
```
Enkelmann, line 374: Could you add to the doc comment, that C/C++ data types are meant here?

Enkelmann (changes requested):

Enkelmann (approved):

PR #174, opened 2021-05-04
Buffer overflow CWE checks

Enkelmann:
This PR adds size hints and bounds checks to stack and heap objects. This allows us to also generate CWE warnings for possible buffer overflows.


src/cwe_checker_lib/src/analysis/pointer_inference/context/trait_impls.rs
```diff
@@ -32,6 +32,36 @@ impl<'a> crate::analysis::forward_interprocedural_fixpoint::Context<'a> for Cont
...
+                    ),
+                ),
+                Def::Store { .. } => (
+                    "CWE787",
+                    format!(
+                        "(Out-of-bounds Write) Memory write at {} may be out of bounds",
+                        def.tid.address
+                    ),
+                ),
+                Def::Assign { .. } => panic!(),
```
mellowCS, line 52: Is is on purpose that you panic without a message?

Enkelmann: Since `state.contains_out_of_bounds_mem_access(...)` always returns `false` for `Def::Assign`, the panic will never execute. So I was too lazy to think of an error message for it.

PR #176, opened 2021-05-05
Better parameter parsing for external Symbols

mellowCS:
This PR contains the following changes:
    - The input parameter for function calls are now parsed from the corresponding Cspec file provided by Ghidra.
    - The Calling Convention struct now contains float and integer parameter separately.


Enkelmann (approved):
Code looks good. Will merge if the CI also agrees.

PR #177, opened 2021-05-05
Dead store removal

Enkelmann:
Add a project normalization pass that removes dead register assignments (i.e. the stored value of the assignment is overwritten before it is read).

This can remove a lot of useless assignments of temporary registers and should result in a speed up of subsequent analyses. In some (probably very rare) cases it can even result in improved accuracy of the pointer inference analysis.


src/cwe_checker_lib/src/analysis/dead_variable_elimination/alive_vars_computation.rs
```diff
@@ -0,0 +1,176 @@
...
+/// modify the set of variables to the ones that are alive before the execution of the `Def` term.
+pub fn update_alive_vars_by_def(alive_variables: &mut HashSet<Variable>, def: &Term<Def>) {
+    match &def.term {
+        Def::Assign { var, value } => {
+            if alive_variables.contains(var) {
+                alive_variables.remove(var);
+                for input_var in value.input_vars() {
+                    alive_variables.insert(input_var.clone());
+                }
+            } else {
```
mellowCS, line 16: Could you not just ommit the else branch?

PR #178, opened 2021-05-07
better Subpiece handling

Enkelmann:
Reworks the implementation of the `Subpiece` and `Piece` operations for the `IntervalDomain` domain to be more exact.


mellowCS (approved):
LGTM

PR #180, opened 2021-05-10
Pointer comparison

Enkelmann:
Handle simple cases of pointer comparisons when specializing via conditional jumps during Pointer Inference algorithm.

### DIFF
--- a/src/cwe_checker_lib/src/analysis/pointer_inference/object_list.rs
+++ b/src/cwe_checker_lib/src/analysis/pointer_inference/object_list.rs
@@ -363,8 +363,8 @@ impl AbstractObjectList {
         }
     }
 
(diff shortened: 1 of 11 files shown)


src/cwe_checker_lib/src/analysis/pointer_inference/object_list.rs
```diff
@@ -374,6 +374,16 @@ impl AbstractObjectList {
             None => Err(()),
         }
     }
+
+    /// Returns `true` if the object corresponding the given ID represents an unique object
```
mellowCS, line 378: There is a 'to' missing after corresponding.

mellowCS (changes requested):
Just a small typo :)

PR #182, opened 2021-05-11
Variable parameter detection

mellowCS:
This PR implements the variable parameter detection for the CWE 78 Check.


src/cwe_checker_lib/src/checkers/cwe_78/state/mod.rs
```diff
@@ -206,10 +210,24 @@ impl State {
...
+    ) {
+        if runtime_memory_image.is_global_memory_address(&constant) {
+            if let Ok(parsed_address) =
+                runtime_memory_image.parse_address_if_recursive(&constant, pointer_byte_size)
+            {
+                match runtime_memory_image.read_string_until_null_terminator(&parsed_address) {
+                    Ok(format_string) => {
+                        self.string_constants.insert(format_string.to_string());
+                    }
+                    Err(e) => panic!("{}", e),
```
Enkelmann, line 227: I am unsure whether this should panic or not. Could this also trigger for cases, where there is no bug on the cwe_checker side, but the input binary is malformed?

src/cwe_checker_lib/src/checkers/cwe_78/state/mod.rs
```diff
@@ -400,6 +449,20 @@ impl State {
         }
     }
 
+    pub fn remove_callee_saved_taint_if_destination_parameter(
```
Enkelmann, line 452: Missing doc comment for the function.

src/cwe_checker_lib/src/utils/binary.rs
```diff
@@ -175,15 +175,51 @@ impl RuntimeMemoryImage {
...
+                } else {
+                    Ok(address.clone())
+                }
+            }
+            Ok(None) => Err(anyhow!(
+                "Writeable address does not guarantee correct content."
+            )),
+            Err(_) => Ok(address.clone()),
+        }
+    }
```
Enkelmann, line 221: Should this function really return `Ok(address.clone())` in some cases? It may confuse the caller if the caller expects the result being the target of address and not address itself.

Enkelmann (changes requested):

src/cwe_checker_lib/src/checkers/cwe_78/context/parameter_detection.rs
```diff
@@ -0,0 +1,370 @@
...
+    /// Checks whether the current symbol is a user input symbol as defined in the symbol configuration.
+    pub fn is_user_input_symbol(&self, symbol: &ExternSymbol) -> bool {
+        self.symbol_maps
+            .user_input_symbol_map
+            .get(&symbol.tid)
+            .is_some()
+    }
+
+    /// Taints the parameters of a non string related extern symbol if it is relevant to the taint analysis.
+    /// To determine whether the symbol is relevant, it is checked if either the arch's return register is tainted
```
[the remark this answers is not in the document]
mellowCS: actually the 'is' should be an 'are'

src/cwe_checker_lib/src/utils/binary.rs
mellowCS: The caller always expects an address to be returned if there is no error. The caller will then resolve the target content. This function's purpose is to parse an recursive address if this is the case. So it could either be (address -> address -> content) or (address -> content). In the first case, the second address is returned.

Enkelmann: But could it happen that in the (address -> content) case the content is misinterpreted as an address?

Enkelmann (approved):

PR #183, opened 2021-05-11
reduce duplicate CWE-416 warnings

Enkelmann:
For cases, where the same dangling pointer is accessed several times in a row, suppress all but the first CWE-416 (use-after-free) warning for it. This reduces the number of duplicate CWE-416 warnings. It should also help in deducing the root cause for a warning, since the remaining warnings are those that are nearest to the corresponding root cause.

### DIFF
--- a/src/cwe_checker_lib/src/analysis/pointer_inference/object.rs
+++ b/src/cwe_checker_lib/src/analysis/pointer_inference/object.rs
@@ -44,7 +44,7 @@ pub struct AbstractObjectInfo {
     /// Tracks whether this may represent more than one actual memory object.
     pub is_unique: bool,
     /// Is the object alive or already destroyed
-    state: Option<ObjectState>,
+    state: ObjectState,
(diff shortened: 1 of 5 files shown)


src/cwe_checker_lib/src/analysis/pointer_inference/object.rs
```diff
@@ -377,6 +385,26 @@ pub enum ObjectState {
     Alive,
     /// The object is dangling, i.e. the memory has been freed already.
     Dangling,
+    /// The state of the object is unknown (due to merging different object states).
+    Unknown,
+    /// The object was referenced in an "use-after-free" or "double-free" CWE-warning.
+    /// This state is meant to be temporary to prevent obvious subsequent CWE-warning with the same root cause.
```
mellowCS, line 391: There is an 's' missing after warning.

mellowCS (changes requested):
Just a small typo

PR #184, opened 2021-05-25
User input detection

mellowCS:
Adds detection of scanf and sscanf calls to the CWE 78 Check.
Also adds artificial arguments to both function signatures since Ghidra does not seem to detect them.


Enkelmann (commented):
Code looks fine to me.
But I would prefer it if we would add the code for the missing `scanf` parameter on the Rust side, not on on the Java side. Mostly because we want to move most of the Java code to the Rust side in the future anyway.

mellowCS:
> Code looks fine to me.
> But I would prefer it if we would add the code for the missing `scanf` parameter on the Rust side, not on on the Java side. Mostly because we want to move most of the Java code to the Rust side in the future anyway.

I'll move it over to the Rust side.

src/cwe_checker_lib/src/pcode/term.rs
```diff
@@ -543,11 +636,12 @@ impl Project {
...
     pub fn into_ir_project(self, binary_base_address: u64) -> IrProject {
         let mut program: Term<IrProgram> = Term {
-            tid: self.program.tid,
+            tid: self.program.tid.clone(),
             term: self
                 .program
                 .term
-                .into_ir_program(binary_base_address, self.stack_pointer_register.size),
+                .clone()
+                .into_ir_program(binary_base_address, &self),
```
Enkelmann, line 644: This cloning is quite expensive (since the IR program should be a very large data structure for large binaries). I think we should try to get rid of the `clone()`.

src/cwe_checker_lib/src/pcode/term.rs
```diff
@@ -655,15 +749,16 @@ impl Project {
...
+            stack_pointer_register: self.stack_pointer_register.clone().into(),
             calling_conventions: self
                 .register_calling_convention
+                .clone()
                 .into_iter()
                 .map(|cconv| cconv.into())
                 .collect(),
             register_list,
-            datatype_properties: self.datatype_properties,
+            datatype_properties: self.datatype_properties.clone(),
```
Enkelmann, line 761: Are the calls to `clone()` here really necessary?

Enkelmann (changes requested):

src/cwe_checker_lib/src/pcode/term.rs
```diff
@@ -392,15 +393,111 @@ pub struct ExternSymbol {
...
+            var: None,
+            location: Some(Expression {
+                mnemonic: ExpressionType::LOAD,
+                input0: Some(Variable {
+                    name: None,
+                    value: None,
+                    address: Some(
+                        address
+                            .to_string()
+                            .repeat(project.stack_pointer_register.size.as_bit_length()),
```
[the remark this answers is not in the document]
Enkelmann: Oh and please double check whether this string should be a hex number or not. :-)

Enkelmann (approved):
Looks good to me now. Will merge if the CI agrees.

PR #185, opened 2021-05-26
Expression propagation

Enkelmann:
When specializing a state because of a conditional jump, we used to go through all the `Def`s in the basic block before the jump (in reverse order) and tried to propagate the knowledge that the jump condition evaluates to `true` if the jump was taken. This unfortunately lead to the addition of too many widening hints to some variables, which in turn lead to too coarse widenings for these variables.

Now we use the following tactic instead:
- We propagate known expressions through a basic block as a normalization pass when creating the project struct
- We simplify many more cases of trivially computable expressions
- We only use the actual condition expression for specializing on conditional jumps. Thanks to the first two steps this expression is in most cases already simplified to the correct comparison expression between physical registers.

### DIFF
--- a/src/cwe_checker_lib/src/analysis/pointer_inference/context/tests.rs
+++ b/src/cwe_checker_lib/src/analysis/pointer_inference/context/tests.rs
@@ -395,97 +395,28 @@ fn specialize_conditional() {
     let mut state = State::new(&register("RSP"), Tid::new("func"));
     state.set_register(&register("RAX"), IntervalDomain::mock(-10, 20).into());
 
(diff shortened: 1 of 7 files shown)


Enkelmann:
@mellowCS It seems that the expression propagation broke the logic for the CWE78-check. The problem seems to be that the check tries to taint a global memory address, which fails and the check subsequently loses track of the taint. Can you think of an easy fix for that? If there is no easy fix we should brainstorm together about other possible solutions.

mellowCS (approved):
LGTM! I've skipped the two CWE 78 tests that fail due to the missing pointer inference state.

PR #186, opened 2021-05-29
Refactor paramter detection

mellowCS:
Refactored the format string parameter detection into its own file in the utils folder. 
This way additional checks can benefit from its functionality.


Enkelmann (approved):
LGTM

PR #188, opened 2021-06-15
Ghidra crash response

mellowCS:
Ghidra now crashes the cwe checker with an error message if something went wrong


src/caller/src/main.rs
```diff
@@ -277,6 +277,14 @@ fn get_project_from_ghidra(file_path: &Path, binary: &[u8]) -> (Project, Vec<Log
             }
         };
 
+        if !String::from_utf8(output.stdout.clone())
+            .unwrap()
```
Enkelmann, line 281: Can you change this unwrap to something like `unwrap_or_else(..)`, where you can tell Rust to `std::process::exit(101)` in the error case?

Enkelmann (approved):
LGTM

PR #189, opened 2021-06-17
add strides to the interval domain

Enkelmann:
Adds tracking of strides to the interval domain.

This improves the exactness of the interval domain and should reduce the number of instances where bounds of loop counter variables are not correctly identified. As a side effect this should also reduce the number of false positive buffer overflow CWE messages, since many of them are caused by incorrect loop counter variables.

### DIFF
--- a/src/cwe_checker_lib/src/abstract_domain/interval.rs
+++ b/src/cwe_checker_lib/src/abstract_domain/interval.rs
@@ -11,13 +11,18 @@ pub use simple_interval::*;
 
 mod bin_ops;
 
-/// An abstract domain representing values in an interval range.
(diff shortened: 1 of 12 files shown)


src/cwe_checker_lib/src/abstract_domain/interval.rs
```diff
@@ -11,13 +11,18 @@ pub use simple_interval::*;
 
 mod bin_ops;
 
-/// An abstract domain representing values in an interval range.
+/// An abstract domain representing values in a interval range with strides and widening hints.
```
mellowCS, line 14: 'an' was correct

mellowCS (approved):
LGTM!

PR #190, opened 2021-06-17
deny missing docs in the CI

Enkelmann:
Missing documentation warnings were already handled as errors when manually running code style checks. Now the CI also handles these as errors.


mellowCS (approved):
LGTM

PR #193, opened 2021-06-30
Error Handling for Parameter Parsing

mellowCS:
The parameter parsing from format strings now allows the developer to log specific errors rather than crashing the analysis in case a format string was externally provided or certain parameters could not be parsed properly.

### DIFF
--- a/src/cwe_checker_lib/src/checkers/cwe_78/context/parameter_detection.rs
+++ b/src/cwe_checker_lib/src/checkers/cwe_78/context/parameter_detection.rs
@@ -103,7 +103,7 @@ impl<'a> Context<'a> {
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


Enkelmann (approved):
Are the `TODO`s for logging of errors planned for a different PR? If yes then I will merge, because the rest of the code looks good to me. You can ignore the clippy errors, since they do not have anything to do with this PR and I have fixed them in PR #189.

mellowCS:
> Are the `TODO`s for logging of errors planned for a different PR? If yes then I will merge, because the rest of the code looks good to me. You can ignore the clippy errors, since they do not have anything to do with this PR and I have fixed them in PR #189.

Yes, they are planned for a future PR.

PR #194, opened 2021-07-01
Mock objects with bytesize

mellowCS:
Mock functions of objects now take a bytesize parameter.
Added two new memory blocks to the memory image mock.


Enkelmann (approved):
LGTM

Issue #192, opened 2021-06-25
thread 'main' panicked at 'called `Result::unwrap()


sh4m2hwz:
$ sudo docker run --rm -v /path/to/binary_file:/path/to fkiecad/cwe_checker path/to -p=CWE78,CWE134,CWE190,CWE215,CWE243,CWE332,CWE367,CWE426,CWE467,CWE476,CWE560,CWE676,CWE782,Memory -j -q -out=f.json 
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 13, kind: PermissionDenied, message: "Permission denied" }', src/cwe_checker_lib/src/utils/log.rs:192:43
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

Enkelmann:
There is no `-out` command line option, there is only `-o` or `--out`. Your invocation is interpreted as `-o ut=f.json` and then the cwe_checker panics, because it has no write access to the file path named `ut=f.json`. Make sure that the output file path is mounted into the docker container to prevent the panic.

Anyway, the unhelpful error message returned to the user is a bug on the cwe_checker side, which we need to fix. So thanks for reporting this!

Enkelmann:
Since the bad error message issue was fixed in #195 and  there was no further response from the issue author, I am assuming that the issue is solved. Feel free to reopen the issue if not.

BusesCanFly:
I've run into this issue when using `--json --out [filename]`. Just `--json` is file, but the addition of the `--out` flag causes:
```rust
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 13, kind: PermissionDenied, message: "Permission denied" }', src/cwe_checker_lib/src/utils/log.rs:192:43
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
It seems the only problem with piping the output to a file instead of using `--out` is that `DEBUG` lines get printed.

---
e.g.
`docker run --rm -v /path/to/bin:/input fkiecad/cwe_checker /input --json > tmp.json; cat tmp.json | grep -v DEBUG > cwe.json; rm tmp.json` is fine while
`docker run --rm -v /path/to/bin:/input fkiecad/cwe_checker /input --json --out cwe.json` errors

Enkelmann:
[...]

PR #195, opened 2021-07-01
better error message on output file write error

Enkelmann:
Fixes an issue raised in #192 where the error message reported to the user does not explain to the user what went wrong.


mellowCS (approved):
LGTM!

PR #198, opened 2021-07-05
Add optional data type param to arg

mellowCS:
Adds an optional data type parameter to the Arg Object. 
This way parameter locations and their size are easier to determine when parsed from a format string.

This PR serves as preparation for a new parameter parsing,
In future not only string parameters but also other data types should be parsed from a format string.

This facilitates the approximation of string abstract domains that are returned from external symbol calls.

### DIFF
--- a/src/cwe_checker_lib/src/utils/arguments.rs
+++ b/src/cwe_checker_lib/src/utils/arguments.rs
@@ -2,6 +2,8 @@
 
 use std::collections::HashMap;
 
+use crate::prelude::*;
+
 use regex::Regex;
 
 use crate::{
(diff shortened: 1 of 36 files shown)


src/cwe_checker_lib/src/utils/arguments.rs
```diff
@@ -209,16 +209,20 @@ pub fn create_string_stack_arg(size: ByteSize, stack_offset: i64) -> Arg {
...
-    })
+    Arg::Register {
+        var: Variable {
+            name: register_name,
+            size,
+            is_temp: false,
+        },
+        data_type: None,
+    }
 }
```
Enkelmann, line 226: Since the function is named create_*string*_register_arg, I would expect it to set the data_type to `Some(Pointer)`.

mellowCS: Fixed.

Enkelmann (approved):
LGTM

PR #200, opened 2021-07-06
Parse all data types from format string

mellowCS:
The argument parser now parses all data types other than long, long long and long double.


Enkelmann (approved):
LGTM

PR #201, opened 2021-07-07
Adapt datatype conversion

mellowCS:
- Added Char type conversion for better char detection in string domains.
- The 'p' format specifier is now interpreted as an integer type instead of pointer.
- Argument promotion for char type is now done in the argument detection.


Enkelmann (approved):
LGTM.

PR #203, opened 2021-07-16
Support analysis of bare metal binaries

Enkelmann:
We now support analyzing bare metal binaries like flash dumps of MCUs.

Note that the analysis is still very experimental and may not yield any reliable or useful output right now.


mellowCS (approved):
LGTM!

PR #204, opened 2021-07-21
Ensure that each block is contained in exactly one subroutine

Enkelmann:
Previously, basic blocks were allowed to be contained in several subroutines (e.g. tail calls generated such cases sometimes). This sometimes resulted in subtle bugs, which were difficult to debug.
This PR implements a normalization pass that duplicates blocks contained in several subroutines so that each subroutine has its own unique copy of the basic block, so that analyses can again assume that each block is only contained in one subroutine.


mellowCS (approved):
LGTM!

PR #205, opened 2021-07-22
IR refactoring

Enkelmann:
Split up the source code files in the intermediate representation module into several files for easier navigation and smaller file sizes.

The PR does not change any functionality, but just moves function and and struct definitions into separate files.


mellowCS (approved):
LGTM

PR #209, opened 2021-08-03
Refactor data domain

Enkelmann:
The `DataDomain` abstract domain is refactored to allow one instance to store both relative targets and absolute values at the same time.

In the previous implementation, the inability to store both relative and absolute values at the same time caused a lot of information loss during the pointer inference analysis, which can be prevented now. Note however, that this PR leaves an in-depth adjustment of the pointer inference analysis to the new `DataDomain` to a future PR. Until then one may actually experience worse results for the pointer inference in some cases. E.g. there seem to be more instances, where the pointer inference algorithm does not stabilize right now.


mellowCS (approved):
LGTM

PR #210, opened 2021-08-04
Add cmdline flag for printing analysis statistics

Enkelmann:
The purpose of the printed statistics is to help in assessing how well the analysis can handle a particular binary.


mellowCS (approved):
LGTM!

PR #211, opened 2021-08-09
Refactor MemRegion for usage with new DataDomain

Enkelmann:
Since the new `DataDomain` does not have a maximal element anymore and uses `Top` as an indicator instead, we have to adjust `MemRegion` accordingly. When merging two `MemRegion` instances, an element only contained in one region gets properly merged with `Top` instead of getting automatically deleted.

As a side effect this further deepens the convergence problems for the Pointer Inference analysis as there are more cases where we track too much information now. So until the whole Pointer Inference analysis is refactored (and a new balance between exactness and convergence is established), the development is not tracked in the master branch but in a separate branch.


mellowCS (approved):
LGTM

PR #212, opened 2021-08-13
improve worklist order for fixpoint algorithm

Enkelmann:
An alternate worklist order for the fixpoint algorithm is implemented. The alternate mode can speed up convergence of the fixpoint algorithm in cases where nodes with a lot of incoming edges have convergence problems otherwise.

The PR also fixes a bug where the priority order for the fixpoint algorithm was reversed incorrectly for all nodes.


mellowCS (approved):
LGTM

PR #213, opened 2021-08-17
MIPS: prevent gp register value loss

Enkelmann:
Until now the pointer inference analysis was prone to lose track of the global pointer value on MIPS binaries. This in turn meant that accesses to global variables could not be tracked. This PR mitigates the issue with the following workarounds:
* On a function call we ensure that the `t9` register is set to the callee address (in accordance with the System V ABI for MIPS). This ensures that the value of `gp` can be correctly computed at the start of a function.
* We also set the `t9` register for entry points.
* After a return from a function call the value of `gp` is usually recovered from the stack. If the stack got invalidated due to unrestricted stack writes this would lose the value of `gp`. So we simply ignore load instructions that would overwrite `gp` with a `Top` value. This workaround only works because Ghidra incorrectly marks `gp` as a callee-saved register (and thus we keep track of the value of `gp` from before the call anyway).

### DIFF
--- a/src/cwe_checker_lib/src/analysis/pointer_inference/context/mod.rs
+++ b/src/cwe_checker_lib/src/analysis/pointer_inference/context/mod.rs
@@ -64,6 +64,47 @@ impl<'a> Context<'a> {
         }
     }
 
+    /// Set the MIPS link register `t9` to the address of the callee TID.
+    ///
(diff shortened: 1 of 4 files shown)


mellowCS (approved):
LGTM

PR #225, opened 2021-09-06
Add automatic pushes to Dockerhub to the CI

Enkelmann:
The Github action for automatically deploying Docker images is extended to also push the Docker images to Dockerhub.  This way we are no longer dependent on the automatic build feature of Dockerhub and can keep Dockerhub as the default container registry for cwe_checker Docker images.

Note that we keep ghcr.io as a secondary container registry so that we can quickly switch to it if Dockerhub deployments should break again in the future.


mellowCS (approved):
LGTM

PR #226, opened 2021-09-08
Correctly classify some buffer overflows as NULL dereferences

Enkelmann:
This PR adds a heuristic to detect cases where a buffer overflow CWE warning is in reality a potential NULL dereference. These cases are now correctly classified as NULL dereference CWE warnings.


mellowCS (approved):
LGTM

PR #227, opened 2021-09-10
add generic parameter objects for entry points

Enkelmann:
Entry points are now generated with a list of generic parameter objects in the *Pointer Inference* analysis. In some cases this drastically increases the number of tracked memory operations, which at the moment outweighs the risk of introducing analysis errors through this change.


mellowCS (approved):
LGTM

PR #228, opened 2021-09-14
WIP: fix error in PointerInference state merging

Enkelmann:
Fixes an error in the merge function for PointerInference states. The function was not yet adjusted to the new DataDomain behavior, which led to incorrectly forgetting knowledge about register values in some cases.


Enkelmann:
Unfortunately the fix seems to lead to runtime explosions for too many binaries in my tests. So we may have to add some heuristics that deliberately forget information to reduce the runtime before we merge the fix. Or wait until the PR for sound abstractions of caller states for the PointerInference is ready. The latter may take a while.

Enkelmann:
This PR is obsolete as the error in question is also fixed in PR #287. Closing.

PR #231, opened 2021-09-23
Minor project struct refactoring

Enkelmann:
Save the extern symbols in a BTreeMap instead of a list in the program struct to simplify lookup operations of extern symbols.


mellowCS (approved):
LGTM

PR #232, opened 2021-09-23
Refactor abstract identifiers

Enkelmann:
The way abstract locations are represented in abstract identifiers is refactored. This should make them a little bit more useful.


mellowCS (approved):
LGTM

PR #233, opened 2021-09-30
Add Cargo.lock

ilkecan:
Hi. I am currently trying to package `cwe_checker` for Nix. While I also think a lock file shouldn't be used for a library; I was expecting to find one in this project because I see `cwe_checker` as an end-product.

I realized the project uses a Cargo workspace, so the Cargo.lock file will be shared by both the binary and the library crate. I wasn't sure what was the best approach with workspaces, so researched a bit and found these discussion:
- <https://stackoverflow.com/questions/62861623/should-cargo-lock-be-committed-when-the-crate-is-both-a-rust-library-and-an-exec>
- <https://old.reddit.com/r/learnrust/comments/htmym0/best_practices_cargolock_in_a_workspace_with_a/>

So the general opinion seem to be to track Cargo.lock file also for workspaces that contains both binary and library crate.

In fact, when you are publishing your crate, library crates won't include Cargo.lock even it exists[^1]. Moreover, even if the lock file is included, cargo won't use it unless you use `--locked`[^2].

[^1]: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html?highlight=publish-lockfile#publish-lockfile
[^2]: Last paragraph of https://doc.rust-lang.org/nightly/cargo/commands/cargo-install.html#description


ilkecan:
I created a PR, but maybe it would be better as an issue. You probably shouldn't let random people update the lock file since it can be used to inject malicious dependencies. https://snyk.io/blog/why-npm-lockfiles-can-be-a-security-blindspot-for-injecting-malicious-modules/

[^1]: https://snyk.io/blog/why-npm-lockfiles-can-be-a-security-blindspot-for-injecting-malicious-modules/

Enkelmann (approved):
I thought a bit about it and yes, you are right, the repository should contain a lockfile.
For your plans of packaging the `cwe_checker` please keep in mind that this is still a research project in early development! I would advise against packaging any version of the master branch, since breaking changes can be introduced there at any time (including those that would break your packaging scripts). Stable version should be OK for packaging.

I manually checked that the lockfile is trustworthy, since the same file is generated on my computer. :-)

PR #278, opened 2022-01-07
Install

m-rtz:
Adds installer for installing cwe_checker. Copies config files at OS dependent locations and searches for a Ghidra installation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,9 +27,8 @@ clean:
 	rm -f -r doc/html
 
 uninstall:
-	rm -f -r ${HOME}/.config/cwe_checker
-	rm -f -r ${HOME}/.local/share/cwe_checker
-	cargo uninstall cwe_checker
+	cargo build -p cwe_checker_install --release
+	./target/release/cwe_checker_install --uninstall
 
 documentation:
 	cargo doc --open --no-deps
--- a/Makefile
+++ b/Makefile
@@ -2,18 +2,9 @@ GHIDRA_PATH =
 
 .PHONY: all clean test uninstall docker
 all:
-	cargo build --release
-ifdef GHIDRA_PATH
-	mkdir -p ${HOME}/.config/cwe_checker
-	cp src/config.json ${HOME}/.config/cwe_checker/config.json
-	cargo install --path src/caller --locked
-	echo "{ \"ghidra_path\": \"${GHIDRA_PATH}\" }" > ${HOME}/.config/cwe_checker/ghidra.json
-	mkdir -p ${HOME}/.local/share/cwe_checker
-	cp -r src/ghidra ${HOME}/.local/share/cwe_checker/ghidra
-else
-	echo "GHIDRA_PATH not specified. Please set it to the path to your local Ghidra installation."
-	false
-endif
+
+	cargo build -p cwe_checker_install --release
+	./target/release/cwe_checker_install ${GHIDRA_PATH}
 
 test:
 	cargo test
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,6 @@ GHIDRA_PATH =
 
 .PHONY: all clean test uninstall docker
 all:
-
 	cargo build -p cwe_checker_install --release
 	./target/release/cwe_checker_install ${GHIDRA_PATH}
 
--- a/src/installer/Cargo.toml
+++ b/src/installer/Cargo.toml
@@ -9,3 +9,4 @@ walkdir = "2"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 anyhow = "1.0" # for easy error types
+structopt = "0.3"
--- a/src/installer/Cargo.toml
+++ b/src/installer/Cargo.toml
@@ -3,10 +3,7 @@ name = "cwe_checker_install"
 version = "0.1.0"
 edition = "2021"
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
-
 [dependencies]
-
 directories = "4.0.1"
 walkdir = "2"
 serde = { version = "1.0", features = ["derive"] }
--- a/src/installer/Cargo.toml
+++ b/src/installer/Cargo.toml
@@ -11,3 +11,4 @@ directories = "4.0.1"
 walkdir = "2"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
+anyhow = "1.0" # for easy error types
--- a/src/installer/Cargo.toml
+++ b/src/installer/Cargo.toml
@@ -1,12 +1,13 @@
 [package]
 name = "cwe_checker_install"
 version = "0.1.0"
-edition = "2018"
+edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
 
-directories = "3.0"
+directories = "4.0.1"
 walkdir = "2"
-fs_extra = "1.2.0"
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
--- a/src/installer/Cargo.toml
+++ b/src/installer/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "cwe_checker_install"
+version = "0.1.0"
+edition = "2018"
+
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -2,18 +2,9 @@ GHIDRA_PATH =
...
-	cp src/config.json ${HOME}/.config/cwe_checker/config.json
-	cargo install --path src/caller --locked
-	echo "{ \"ghidra_path\": \"${GHIDRA_PATH}\" }" > ${HOME}/.config/cwe_checker/ghidra.json
-	mkdir -p ${HOME}/.local/share/cwe_checker
-	cp -r src/ghidra ${HOME}/.local/share/cwe_checker/ghidra
-else
-	echo "GHIDRA_PATH not specified. Please set it to the path to your local Ghidra installation."
-	false
-endif
+
```
Enkelmann, line 5: We should remove this empty line.

src/installer/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "cwe_checker_install"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
Enkelmann, line 6: Just to nitpick a little: This comment can be removed. ;-)

src/installer/src/main.rs
```diff
@@ -0,0 +1,232 @@
...
+            return Some(hit);
+        }
+    }
+    None
+}
+
+/// Determines Ghidra versions and provides selection interface for the user.
+fn select_ghidra_version(mut ghidra_locations: Vec<PathBuf>) -> Result<PathBuf> {
+    let mut i = 0;
+    while i < ghidra_locations.len() {
```
Enkelmann, line 99: This should be a for-loop instead of a while-loop with manually incremented loop counter.

src/installer/src/main.rs
```diff
@@ -0,0 +1,232 @@
...
+                i += 1;
+            }
+
+            Err(_) => {
+                ghidra_locations.remove(i);
+            }
+        }
+    }
+    if ghidra_locations.is_empty() {
+        return Err(Error::new(ErrorKind::NotFound, "Ghidra not found").into());
```
Enkelmann, line 127: Since we are using the anyhow crate, we can write this more concise as
```rust
return Err(anyhow!("Ghidra not found."));
```
This is true for all error messages you use.

src/installer/src/main.rs
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    get_user_choice(&ghidra_locations)
+}
+
+/// Determines Ghidra versions and provides selection interface for the user.
+fn get_user_choice(ghidra_locations: &[PathBuf]) -> Result<PathBuf> {
```
Enkelmann, line 134: Such a choice should also contain an option to abort the installation (which it does not currently).

src/installer/src/main.rs
```diff
@@ -0,0 +1,232 @@
...
+
+    std::fs::write(location.join("ghidra.json"), serde_json::to_string(&conf)?)?;
+    Ok(())
+}
+
+/// Runs Cargo install to install cwe_checker.
+fn install_cwe_checker() -> Result<()> {
+    std::process::Command::new("cargo")
+        .args(["install", "--path", "src/caller", "--locked"])
+        .status()?;
```
Enkelmann, line 167: When I tried the install script in my Windows VM, this failed (because of something unrelated to your PR), but the installation script did not stop. So apparently the return value of this function was still `Ok(())`. We should investigate why.

src/installer/src/main.rs
```diff
@@ -0,0 +1,232 @@
...
+    Ok(())
+}
+
+// Recursive copy of files and directories.
```
Enkelmann, line 171: Documentation comments need three slashes, not two. ;-)

src/installer/src/main.rs
```diff
@@ -0,0 +1,232 @@
...
+        if ty.is_dir() {
+            copy_dir_all(entry.path(), dst.as_ref().join(entry.file_name()))?;
+        } else {
+            fs::copy(entry.path(), dst.as_ref().join(entry.file_name()))?;
+        }
+    }
+    Ok(())
+}
+
+// Copy src/ghidra to provided location.
```
Enkelmann, line 186: Documentation comments need three slashes, not two. ;-)

src/installer/src/main.rs
```diff
@@ -0,0 +1,232 @@
...
+    if args.len() == 2 {
+        create_ghidra_json(cwe_checker_conf_dir.config_dir(), PathBuf::from(&args[1]))?;
+    } else if cwe_checker_conf_dir
+        .config_dir()
+        .join("ghidra.json")
+        .exists()
+    {
+        println!(
+            "found ghidra.json at {}, keeping it.",
+            cwe_checker_conf_dir.config_dir().display()
```
Enkelmann, line 214: I think the `.display()` at the end should be redundant.

src/installer/src/main.rs
```diff
@@ -0,0 +1,232 @@
...
+        match find_ghidra() {
+            Ok(ghidra_location) => {
+                create_ghidra_json(cwe_checker_conf_dir.config_dir(), ghidra_location)?;
+            }
+            Err(err) => return Err(err),
+        }
+    }
+
+    println!("installing CWE-Checker...");
+    install_cwe_checker()?;
```
Enkelmann, line 227: We should make sure that we do not needlessly copy configuration files if the compilation step fails. So we should do the compilation step before any config file creation (but after selection of the ghidra path).

src/installer/src/main.rs
```diff
@@ -0,0 +1,232 @@
...
+
+    println!("copy Ghidra Plugin...");
```
Enkelmann, line 229: It would be nice to include the actual path to where it is copied to in the log message. Same goes for the other configuration files.

Enkelmann (changes requested):
Some general observations:
- When I tried this, it found the same installation folder of Ghidra twice for me. So we should add some deduplication step before the choice dialog for the user.
- We also have to rustify the uninstall step now. Else the paths in the makefile are not the actual installation paths for the configuration files.

src/installer/src/main.rs
m-rtz: It seems like `Path` doesn't implement `std::fmt::Display`.

src/installer/src/main.rs
Enkelmann: I identified the problem: `.status()?` checks only that the command finished by unwrapping the exit code of the command. But it does not actually check whether the exit code is zero or not, i.e. we do not check whether it was a success or not. So we also have to check the return value of `.status()?` to make sure that the `cargo install` command was successful.

src/installer/src/main.rs
```diff
@@ -0,0 +1,271 @@
...
+    std::fs::remove_dir_all(conf_dir)?;
+    std::fs::remove_dir_all(data_dir)?;
+    std::process::Command::new("cargo")
+        .args(["uninstall", "cwe_checker"])
+        .status()?;
+    Ok(())
+}
+
+fn main() -> Result<()> {
+    let cwe_checker_conf_dir = ProjectDirs::from("", "", "cwe_checker").unwrap();
```
Enkelmann, line 218: Just to nitpick a little: `cwe_checker_conf_dir` contains both the config dir and the data dir for the cwe_checker, so the name of the variable is slightly misleading.

src/installer/src/main.rs
```diff
@@ -0,0 +1,271 @@
...
+                    Ok(ghidra_location) => {
+                        create_ghidra_json(cwe_checker_conf_dir.config_dir(), ghidra_location)?;
+                    }
+                    Err(err) => return Err(err),
+                }
+            }
+        },
+    }
+
+    println!("installing CWE-Checker...");
```
Enkelmann, line 256: The official naming convention is `cwe_checker` and not `CWE-Checker`. ;-)

Enkelmann (changes requested):
This is almost ready to be merged. :-) Apart from the Windows-issue (which is actually not Windows-related at all) everything worked as intended in my tests.

Enkelmann (approved):
Code looks good now. :-) Merging once the CI also approves.

PR #285, opened 2022-02-07
Test reorganization

m-rtz:
Tests for calling conventions edited for meeting System V ABI/ARM32 ABI and Ghidras output.

### DIFF
--- a/src/cwe_checker_lib/src/intermediate_representation/sub.rs
+++ b/src/cwe_checker_lib/src/intermediate_representation/sub.rs
@@ -226,9 +226,21 @@ mod tests {
                     Variable::mock("R8", 8),
                     Variable::mock("R9", 8),
                 ],
-                float_parameter_register: vec![Expression::Var(Variable::mock("XMM0", 16))],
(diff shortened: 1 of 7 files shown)


Enkelmann (changes requested):

src/cwe_checker_lib/src/intermediate_representation/sub.rs
```diff
@@ -213,20 +213,17 @@ mod tests {
             }
         }
     }
-    /// Returns model for floating points for argument passing in x64
-    fn create_x64_float_register_subpice(name: &str) -> Expression {
+    /// Wrapper for subpiece to model float register for arument passing
```
Enkelmann, line 216: New typo: `arument`

Enkelmann (changes requested):
You forgot to fix the ARM-32 integer parameter registers and one typo. The rest looks good. :-)

Enkelmann (approved):
Code looks good now. :+1: I will merge it if the CI also agrees.

PR #291, opened 2022-02-21
Testreorganisation project struct

m-rtz:


src/cwe_checker_lib/src/intermediate_representation/program.rs
```diff
@@ -48,5 +48,49 @@ mod tests {
...
+            let free = ExternSymbol {
+                tid: Tid::new("free"),
+                addresses: vec![],
+                name: "free".to_string(),
+                calling_convention: None,
+                parameters: vec![],
+                return_values: vec![],
+                no_return: false,
+                has_var_args: false,
+            };
```
Enkelmann, line 72: malloc and free should have parameters and return values as expected from the corresponding libc functions in the mocked program. This probably means that we need 2 versions, one for x64 and one for ARM, because of the different param/return registers.

src/cwe_checker_lib/src/intermediate_representation/project.rs
```diff
@@ -249,21 +249,68 @@ mod tests {
...
+                .chain(none_cconv_16byte_register)
+                .chain(callee_saved_register);
+
+            Project {
+                program: Term {
+                    tid: Tid::new("program_tid"),
+                    term: Program::mock(),
+                },
+                cpu_architecture: "arm32".to_string(),
+                stack_pointer_register: Variable::mock("r13", 4u64),
```
Enkelmann, line 307: I think Ghidra names the stack pointer register `sp` instead of `r13`. Can you double-check and if necessary change this?

Enkelmann (changes requested):
- Could you remove the old `DatatypeProperties::mock()` function? It should be replaced by your new `mock_x64()` variant.
- `Program::mock_empty()` should also be obsolete now.

m-rtz:
> * Could you remove the old `DatatypeProperties::mock()` function? It should be replaced by your new `mock_x64()` variant.

This function is still used in `src/cwe_checker_lib/src/analysis/pointer_inference/context/tests.rs`. This requires some other refactoring for the next iteration.

src/cwe_checker_lib/src/intermediate_representation/program.rs
```diff
@@ -37,16 +37,87 @@ impl Program {
...
+        /// Returns Program with malloc, free and other_function
+        pub fn mock_arm32() -> Program {
+            // Creates arm32 program by altering x64 program
+            let mut prog = Program::mock_x64();
+            for symbol in prog.extern_symbols.values_mut() {
+                if symbol.name == "malloc".to_string() {
+                    symbol.parameters = vec![Arg::from_var(
+                        CallingConvention::mock_arm32().integer_parameter_register[0].clone(),
+                        Some(Datatype::Integer),
+                    )];
```
Enkelmann, line 107: While your code works as intended, its structure could backfire in the future if we change something in the `mock_x64` function without realizing that it also affects the `mock_arm32` function. One possible better structure would be using a helper function for creating the extern symbols which gets the calling convention as a parameter.

Enkelmann (changes requested):
I still have one change request related to clean code, but the rest look good now. :-)

Enkelmann (approved):
LGTM

PR #317, opened 2022-04-25
Stack alignment analysis

m-rtz:


src/cwe_checker_lib/src/analysis/stack_alignment_substitution/mod.rs
```diff
@@ -0,0 +1,401 @@
...
+        "x86_64" => 16,
+        "arm32" => 4,
+        _ => 0,
+    };
+
+    let journaled_sp: &mut u32 = &mut 10000000; // 128-Byte aligned
+    let mut log: Vec<LogMessage> = vec![];
+
+    for sub in project.program.term.subs.values_mut() {
+        for blk in sub.term.blocks.iter_mut() {
```
Enkelmann, line 77: Since we do not track the stack pointer value over several basic blocks of a function, our analysis can at most be correct for the first basic block of a function. Thus we should only substitute in the first block of a function.

src/cwe_checker_lib/src/analysis/stack_alignment_substitution/mod.rs
```diff
@@ -0,0 +1,401 @@
...
+            for def in blk.term.defs.iter_mut() {
+                if let Def::Assign { var, value } = &mut def.term {
+                    if *var == project.stack_pointer_register {
+                        if let Expression::BinOp { op, lhs, rhs } = value {
```
Enkelmann, line 81: I think this `if` needs an `else` for detecting cases where something unexpected happens with the stack pointer.

src/cwe_checker_lib/src/analysis/stack_alignment_substitution/mod.rs
```diff
@@ -0,0 +1,401 @@
...
+                            if *lhs
+                                == Box::new(Expression::Var(project.stack_pointer_register.clone()))
```
Enkelmann, line 83: It is not guaranteed that the stack pointer is the left operand. It may also be the right operand. Also: We should handle the `else` case here since it would denote a case where something unexpected happens with the stack pointer.

Enkelmann (changes requested):
The analysis does not seem to stop after encountering an unsupported operation. But it should stop in such a case, since it means that we have lost the knowledge over the correct stack pointer value and thus further substitutions are probably wrong.

Enkelmann (changes requested):
There is also still one issue from the first review that you missed. This time I also reviewed the unit tests and they look very good!

Edit: The merge conflicts now are due to another PR I merged today (sorry). But they should be easily resolvable if you merge the master branch into this branch.

Enkelmann (changes requested):
One Clippy issue to fix and you forgot to fix one typo. Apart from that everything looks ready to be merged. :-)

Enkelmann (approved):
Now everything looks fine! If the CI also agrees, I will merge it.

PR #336, opened 2022-06-13
added Pointer Inference results as basis for overflow detection.

m-rtz:
Added Pointer Inference results for detecting CWE190.

### DIFF
--- a/src/cwe_checker_lib/src/abstract_domain/interval.rs
+++ b/src/cwe_checker_lib/src/abstract_domain/interval.rs
@@ -373,6 +373,11 @@ impl IntervalDomain {
             widening_delay,
         }
     }
+
+    /// Returns the contained interval.
+    pub fn get_interval(&self) -> &Interval {
+        &self.interval
+    }
 }
 
 impl SpecializeByConditional for IntervalDomain {
--- a/src/cwe_checker_lib/src/abstract_domain/interval.rs
+++ b/src/cwe_checker_lib/src/abstract_domain/interval.rs
@@ -373,11 +373,6 @@ impl IntervalDomain {
             widening_delay,
         }
     }
-
-    /// Returns the contained interval.
-    pub fn get_interval(&self) -> &Interval {
-        &self.interval
-    }
 }
 
 impl SpecializeByConditional for IntervalDomain {
--- a/src/cwe_checker_lib/src/checkers/cwe_190.rs
+++ b/src/cwe_checker_lib/src/checkers/cwe_190.rs
@@ -122,7 +122,7 @@ fn calloc_parm_mul_is_top(pir: &PointerInference, jmp_tid: &Tid, parms: Vec<&Arg
         pir.eval_parameter_arg_at_call(jmp_tid, parms[1]),
     ) {
         nmeb.bin_op(BinOpType::IntMult, &size);
-        return contains_only_non_top_absolute_value(&nmeb);
+        return !contains_only_non_top_absolute_value(&nmeb);
     }
 
     false
--- a/src/cwe_checker_lib/src/checkers/cwe_190.rs
+++ b/src/cwe_checker_lib/src/checkers/cwe_190.rs
@@ -1,6 +1,6 @@
 //! This module implements a check for CWE-190: Integer overflow or wraparound.
 //!
-//! An integer overflow can lead to undefined behaviour and is especially dangerous
+//! An integer overflow can lead to undefined behavior and is especially dangerous
 //! in conjunction with memory management functions.
 //!
 //! See <https://cwe.mitre.org/data/definitions/190.html> for a detailed description.
@@ -20,7 +20,7 @@
 //! - There is no check whether the result of the multiplication is actually used
 //!   as input to the function call. However, this does not seem to generate a lot
 //!   of false positives in practice.
-//! - Values that are not absolute e.g. user controled or depend on other values.
+//! - Values that are not absolute e.g. user controlled or depend on other values.
 //!
 //! ## False Negatives
 //!
@@ -28,6 +28,7 @@
 //! from the CWE190 symbol list.
 //! - All integer overflows caused by addition or subtraction.
 
+use crate::abstract_domain::AbstractDomain;
 use crate::abstract_domain::RegisterDomain;
 use crate::analysis::pointer_inference::*;
 use crate::analysis::vsa_results::*;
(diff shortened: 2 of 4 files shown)


src/cwe_checker_lib/src/checkers/cwe_190.rs
```diff
@@ -10,7 +10,8 @@
 //! For each call to a function from the CWE190 symbol list we check whether the
 //! basic block directly before the call contains a multiplication instruction.
 //! If one is found, the call gets flagged as a CWE hit, as there is no overflow
-//! check corresponding to the multiplication before the call. The default CWE190
+//! check corresponding to the multiplication before the call and the Pointer Inference
+//! returns argument intervals containing top values. The default CWE190
```
Enkelmann, line 14: Well, we now only flag something as a CWE if a multiplication is found *and* we cannot guarantee that no overflow occurred due to Pointer Inference results. Your formulation still suggests that we only check for the multiplication instruction.

src/cwe_checker_lib/src/checkers/cwe_190.rs
```diff
@@ -19,15 +20,16 @@
 //! - There is no check whether the result of the multiplication is actually used
 //!   as input to the function call. However, this does not seem to generate a lot
 //!   of false positives in practice.
-//! - There is no value set analysis in place to determine whether an overflow is
-//!   possible or not at the specific instruction.
+//! - Relative values are always generate a hit.
```
Enkelmann, line 23: The reader might not know about the intrinsics of the cwe_checker and thus might not know what a relative value is in this context.

src/cwe_checker_lib/src/checkers/cwe_190.rs
```diff
@@ -97,6 +99,24 @@ fn generate_cwe_warning(callsite: &Tid, called_symbol: &ExternSymbol) -> CweWarn
         .symbols(vec![called_symbol.name.clone()])
 }
 
+/// Determines if the argument for the call contains relative intervals or top values for absolute intervals.
+fn contains_top_value(pir: &PointerInference, jmp_tid: &Tid, parms: Vec<&Arg>) -> bool {
+    for arg in parms {
+        let value = pir.eval_parameter_arg_at_call(jmp_tid, arg).unwrap();
+        dbg!(&value);
```
Enkelmann, line 106: You forgot to remove a debug print.

src/cwe_checker_lib/src/checkers/cwe_190.rs
```diff
@@ -97,6 +99,24 @@ fn generate_cwe_warning(callsite: &Tid, called_symbol: &ExternSymbol) -> CweWarn
         .symbols(vec![called_symbol.name.clone()])
 }
 
+/// Determines if the argument for the call contains relative intervals or top values for absolute intervals.
+fn contains_top_value(pir: &PointerInference, jmp_tid: &Tid, parms: Vec<&Arg>) -> bool {
+    for arg in parms {
+        let value = pir.eval_parameter_arg_at_call(jmp_tid, arg).unwrap();
```
Enkelmann, line 105: We cannot call `unwrap()` here! `eval_parameter_arg_at_call` may return `None` if there is no Pointer Inference state at the call (this can happen e.g. for dead code). But calling `unwrap()` in such a case would crash the cwe_checker which is the wrong response in this case.

src/cwe_checker_lib/src/checkers/cwe_190.rs
```diff
@@ -97,6 +99,24 @@ fn generate_cwe_warning(callsite: &Tid, called_symbol: &ExternSymbol) -> CweWarn
...
+        dbg!(&value);
+        if let Some(inter_dom) = value.get_absolute_value() {
+            if inter_dom.get_interval().is_top() {
+                return true;
+            }
+        }
+        if !value.get_relative_values().is_empty() {
+            return true;
+        }
```
Enkelmann, line 114: You can use the method `value.get_if_absolute_value()` to get the contained absolute values and make sure that no other type of values are contained in the domain at the same time.

Also, the interval domain implements the `AbstractDomain` trait, i.e. you can simply write `inter_dom.is_top()` without unwrapping the contained interval first.

src/cwe_checker_lib/src/abstract_domain/interval.rs
```diff
@@ -373,6 +373,11 @@ impl IntervalDomain {
             widening_delay,
         }
     }
+
+    /// Returns the contained interval.
+    pub fn get_interval(&self) -> &Interval {
+        &self.interval
+    }
 }
```
Enkelmann, line 381: This method turns out to be unnecessary. See the other review comments.

src/cwe_checker_lib/src/checkers/cwe_190.rs
```diff
@@ -105,13 +125,25 @@ pub fn check_cwe(
...
     for sub in project.program.term.subs.values() {
         for (block, jump, symbol) in get_callsites(sub, &symbol_map) {
             if block_contains_multiplication(block) {
-                cwe_warnings.push(generate_cwe_warning(&jump.tid, symbol));
+                let parms = match symbol.name.as_str() {
+                    "malloc" => vec![&symbol.parameters[0]],
+                    "xmalloc" => vec![&symbol.parameters[0]],
+                    "calloc" => vec![&symbol.parameters[0], &symbol.parameters[1]],
+                    "realloc" => vec![&symbol.parameters[1]],
+                    _ => vec![&symbol.parameters[0]],
```
Enkelmann, line 141: The general case should be to check all parameters for top values, since users might add other functions to the list where we do not know which parameters to check.. We can look up the list of all parameters through `project.program.term.extern_symbols`. And to nitpick a little more, we only need special cases for `realloc` (like you implemented it) and for `calloc` (where we should additionally check that the product of the two parameters does not overflow). But `malloc` and `xmalloc` do not need special cases. ;-)

Enkelmann (changes requested):
Can you check why the acceptance tests fail now? If they contain cases where integer overflows are actually impossible and the analysis is now smart enough to discern that we may have to adjust the acceptance tests.

src/cwe_checker_lib/src/checkers/cwe_190.rs
```diff
@@ -97,6 +101,35 @@ fn generate_cwe_warning(callsite: &Tid, called_symbol: &ExternSymbol) -> CweWarn
...
+}
+
+/// Checks if the multiplication of element count and size parameters result in an overflow.
+fn calloc_parm_mul_is_top(pir: &PointerInference, jmp_tid: &Tid, parms: Vec<&Arg>) -> bool {
+    if let (Some(nmeb), Some(size)) = (
+        pir.eval_parameter_arg_at_call(jmp_tid, parms[0]),
+        pir.eval_parameter_arg_at_call(jmp_tid, parms[1]),
+    ) {
+        nmeb.bin_op(BinOpType::IntMult, &size);
+        nmeb.contains_top()
```
Enkelmann, line 127: Here we should use the same check as in the `contains_top_value` function, i.e. unwrap the absolute value with `nmeb.get_if_absolute_value()` to check the contained absolute value for being a `Top` value.

Enkelmann (changes requested):
One small issue with the `calloc` handling (I think I missed this in my last review). The rest looks good. :)

Enkelmann (approved):
Now everything looks good to me. :-)

PR #341, opened 2022-07-07
added Display trait for Expression and Operants

m-rtz:
Implemented the Display Trait for Expression by using somewhat intuitive symbols.

### DIFF
--- a/src/cwe_checker_lib/src/intermediate_representation/def.rs
+++ b/src/cwe_checker_lib/src/intermediate_representation/def.rs
@@ -1,3 +1,5 @@
+use std::fmt;
+
 use super::{CastOpType, Expression, Variable};
 use crate::prelude::*;
 
@@ -82,6 +84,16 @@ impl Term<Def> {
                 address.substitute_input_var(input_var, replace_with_expression);
                 value.substitute_input_var(input_var, replace_with_expression);
             }
+        }
+    }
+}
+
+impl fmt::Display for Def {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Def::Load { var, address } => write!(f, "{} = @{}", var, address),
+            Def::Store { address, value } => write!(f, "@{} = {}", address, value),
+            Def::Assign { var, value } => write!(f, "{} = {}", var, value),
         }
     }
 }
--- a/src/cwe_checker_lib/src/intermediate_representation/def.rs
+++ b/src/cwe_checker_lib/src/intermediate_representation/def.rs
@@ -91,8 +91,8 @@ impl Term<Def> {
 impl fmt::Display for Def {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            Def::Load { var, address } => write!(f, "{} = @{}", var, address),
-            Def::Store { address, value } => write!(f, "@{} = {}", address, value),
+            Def::Load { var, address } => write!(f, "{} := Load from {}", var, address),
+            Def::Store { address, value } => write!(f, "Store at {} := {}", address, value),
             Def::Assign { var, value } => write!(f, "{} = {}", var, value),
         }
     }
--- a/src/cwe_checker_lib/src/intermediate_representation/expression.rs
+++ b/src/cwe_checker_lib/src/intermediate_representation/expression.rs
@@ -439,9 +439,16 @@ impl fmt::Display for Expression {
             Expression::Const(c) => {
                 write!(f, "0x{:016x}:i{}", c, c.bytesize().as_bit_length())
             }
-            Expression::BinOp { op, lhs, rhs } => write!(f, "({} {} {})", lhs, op, rhs),
-            Expression::UnOp { op, arg } => write!(f, "({}({}))", op, arg),
-            Expression::Cast { op, size: _, arg } => write!(f, "({}({}))", op, arg),
+            Expression::BinOp { op, lhs, rhs } => match op {
+                BinOpType::IntMult
+                | BinOpType::IntDiv
+                | BinOpType::IntRem
+                | BinOpType::FloatMult
+                | BinOpType::FloatDiv => write!(f, "{} {} {}", lhs, op, rhs),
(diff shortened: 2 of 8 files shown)


Enkelmann (changes requested):

src/cwe_checker_lib/src/intermediate_representation/expression/tests.rs
```diff
@@ -539,3 +539,22 @@ fn processing_sub_registers() {
...
+    };
+    let expr = expr
+        .cast(CastOpType::IntSExt)
+        .un_op(UnOpType::FloatCeil)
+        .subpiece(ByteSize(0), ByteSize(20));
+
+    assert_eq!(
+        "((FloatCeil((IntSExt((¬((2 + (EAX:32 - ECX:32))))))))₀₋₎",
+        format!("{}", expr)
+    );
```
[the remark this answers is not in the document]
m-rtz: Do you mean for arithmetic to distinguish e.g. if multiplication is in unnecessary brackets? Or do you suggest a different style for e.g. casts?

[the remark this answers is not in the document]
Enkelmann: Well, both. Most importantly, it seems that we print each pair of brackets twice instead of once. E.g. `FloatCeil` puts brackets around its argument expression, but the argument also puts brackets around itself.

src/cwe_checker_lib/src/intermediate_representation/def.rs
```diff
@@ -86,6 +88,16 @@ impl Term<Def> {
     }
 }
 
+impl fmt::Display for Def {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Def::Load { var, address } => write!(f, "{} = @{}", var, address),
+            Def::Store { address, value } => write!(f, "@{} = {}", address, value),
```
Enkelmann, line 95: I prefer C-like syntax, i.e. `*{}` instead of `@{}` in Load and Store instructions. Maybe we should also use `:=` instead of `=` to make it explicitly clear that the right side is assigned to the left side. Although this would not be C-like syntax...

src/cwe_checker_lib/src/intermediate_representation/expression.rs
```diff
@@ -431,5 +432,79 @@ impl Expression {
...
+                description,
+                size: _,
+            } => write!(f, "{}", description),
+            Expression::Subpiece {
+                low_byte,
+                size,
+                arg,
+            } => {
+                if let (Ok(start), Ok(end)) = (u32::try_from(low_byte.0), u32::try_from(size.0)) {
+                    write!(f, "({})[{}₋{}]", arg, start, end)
```
Enkelmann, line 462: For some reason the character you use as hyphen `₋` is not the normal `-` (minus) character. I think we should use the normal `-`.

src/cwe_checker_lib/src/intermediate_representation/expression.rs
```diff
@@ -431,5 +432,79 @@ impl Expression {
...
+            BinOpType::IntSub => write!(f, "-"),
+            BinOpType::IntXOr => write!(f, "^"),
+            BinOpType::IntAnd => write!(f, "&"),
+            BinOpType::IntOr => write!(f, "|"),
+            BinOpType::IntLeft => write!(f, "<<"),
+            BinOpType::IntRight => write!(f, ">>"),
+            BinOpType::IntMult => write!(f, "*"),
+            BinOpType::IntDiv => write!(f, "/"),
+            BinOpType::IntRem => write!(f, "%"),
+            _ => write!(f, "{:?}", self),
```
Enkelmann, line 490: Could you add the `BoolAnd` and `BoolOr` operations here? If we use `&&` and `||` for them there is no risk of confusing them with the byte-wise versions.

src/cwe_checker_lib/src/intermediate_representation/expression.rs
```diff
@@ -431,5 +432,79 @@ impl Expression {
...
+        }
+    }
+}
+
+impl fmt::Display for UnOpType {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            UnOpType::IntNegate => write!(f, "¬"),
+            _ => write!(f, "{:?}", self),
```
Enkelmann, line 499: We should also add the `BoolNegate` operation here. I would actually use `¬` for it, although then we should use something different for `IntNegate`. Maybe `~` (tilde)? This would be the C-operator for bitwise negation.

src/cwe_checker_lib/src/intermediate_representation/jmp.rs
```diff
@@ -58,3 +60,35 @@ pub enum Jmp {
         return_: Option<Tid>,
     },
 }
+
+impl fmt::Display for Jmp {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Jmp::Branch(tid) => write!(f, "-> {}", tid),
+            Jmp::BranchInd(expr) => write!(f, "-> {}", expr),
+            Jmp::CBranch { target, condition } => write!(f, "if {} -> {}", condition, target),
```
Enkelmann, line 69: I would prefer writing `jump to` instead of `->`, since it is easier to understand for someone seeing it the first time. At least for me it was a little confusing at first until I understood what the `->` meant.

Enkelmann (changes requested):
Looks good! And the number of brackets is now also on an acceptable level. I still have some small style issues I would like to change. Apart from them the PR is ready to be merged.

src/cwe_checker_lib/src/intermediate_representation/def.rs
Enkelmann: Another possibility would be `... := Load from ...` for loads and `Store at ... := ...` for stores. This variant would be easier to spot when scrolling over larger amounts of code.

Enkelmann (approved):
Thanks! Will merge if the CI also approves.

PR #347, opened 2022-07-25
Cwe789

m-rtz:
Implemented CWE789: Memory Allocation with Excessive Size Value

The check uses thresholds for stack memory allocations and heap memory allocation defined in config.json. For the stack allocation case, every assignment onto the stackpointer is checked, if it exceeds the threshold.
For the heap memory case the functions xmalloc, malloc, realloc and calloc are predefined and their arguments are checked if they exceed the threshold.

### DIFF
--- a/src/cwe_checker_lib/src/checkers/cwe_789.rs
+++ b/src/cwe_checker_lib/src/checkers/cwe_789.rs
@@ -1,6 +1,6 @@
 //! This module implements a check for CWE-789: Memory Allocation with Excessive Size Value.
 //!
-//! Only stack memory allocation is covered in this module.
+//! Stack memory allocation and function calls like malloc are covered in this module.
 //! Excessive allocation of memory might destabilize programs on machines with limited resources.
 //!
 //! See <https://cwe.mitre.org/data/definitions/789.html> for a detailed description.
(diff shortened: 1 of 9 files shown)


m-rtz:
If an warning is generated for the stack allocation case, it seems like a new warning is generated on every following assignment onto the stack pointer, including calls. This might lead to a large amount of warnings.

src/cwe_checker_lib/src/checkers/cwe_789.rs
```diff
@@ -0,0 +1,176 @@
...
+            let assign_on_sp: Vec<&Term<Def>> = blk
+                .term
+                .defs
+                .iter()
+                .filter(|x| is_assign_on_sp(&x.term, &project.stack_pointer_register))
+                .collect();
+            for assign in assign_on_sp {
+                if let Some(interval) = analysis_results
+                    .pointer_inference
+                    .unwrap()
```
Enkelmann, line 163: You have already stored the result of `analysis_results.pointer_inference.unwrap()` in the `pir` variable.

src/cwe_checker_lib/src/checkers/cwe_789.rs
```diff
@@ -0,0 +1,176 @@
...
+                    .eval_value_at_def(&assign.tid)
+                {
+                    if exceeds_threshold_on_stack(interval, config.stack_threshold) {
+                        cwe_warnings.push(generate_cwe_warning(&assign.tid, true));
+                    }
+                }
+            }
+        }
+    }
+    cwe_warnings.dedup();
```
Enkelmann, line 173: This `dedup()` call unfortunately does not help us here. One possible way to only produce at most one stack-related CWE warning per function is to add a `break` (or `continue` the outer loop) after the first CWE warning is generated for a function.

Enkelmann (changes requested):
The check looks good!

Apart from reducing the number of duplicates for stack-related CWE warnings this check also needs an acceptance check!

Have you tried this on some firmware binaries? I fear that the standard `heap_threshold` in the `config.json` might be too small. The `stack_threshold` on the other hand should be smaller in my opinion. But we should test that on some real world examples before deciding the standard threshold values.

src/cwe_checker_lib/src/checkers/cwe_789.rs
```diff
@@ -0,0 +1,176 @@
...
+fn multiply_args_for_calloc(
+    pir: &PointerInference,
+    jmp_tid: &Tid,
+    parms: Vec<&Arg>,
+) -> Option<DataDomain<IntervalDomain>> {
+    if let (Some(nmeb), Some(size)) = (
+        pir.eval_parameter_arg_at_call(jmp_tid, parms[0]),
+        pir.eval_parameter_arg_at_call(jmp_tid, parms[1]),
+    ) {
+        nmeb.bin_op(BinOpType::IntMult, &size);
```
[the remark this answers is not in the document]
m-rtz: Good Hint! Using this I discovered the same bug in CWE190.

m-rtz:
> The check looks good!
> 
> Apart from reducing the number of duplicates for stack-related CWE warnings this check also needs an acceptance check!
> 
> Have you tried this on some firmware binaries? I fear that the standard `heap_threshold` in the `config.json` might be too small. The `stack_threshold` on the other hand should be smaller in my opinion. But we should test that on some real world examples before deciding the standard threshold values.

I evaluated the check on some firmware binaries and the current values: 1MB for heap allocations and 7500B for stack allocation seems to be a starting point for this check.

PR #356, opened 2022-10-17
Expression propagation rewrite

m-rtz:
Added a re-implementation of the expression propagation routine for preprocessing assignments. This is now inter-bassic-block wide and uses a fixpoint algorithm. Small refactoring of `Blk::propagate_input_expressions()`.


src/cwe_checker_lib/src/analysis/expression_propagation/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+                .neighbors_directed(node, petgraph::Incoming)
+                .next()
+                .is_none()
+            {
+                computation.set_node_value(node, NodeValue::Value(HashMap::new()));
+            }
+        }
+    }
+
+    computation.compute_with_max_steps(100);
```
Enkelmann, line 166: We also need to set the `BlkStart` values of blocks that are entry points into functions! Else we will miss most basic blocks in the fixpoint computation.

src/cwe_checker_lib/src/analysis/expression_propagation/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+            first_block.merge_def_assignments_to_same_var();
+            first_block.propagate_input_expressions(None);
+        }
+        for block in blocks {
+            block.merge_def_assignments_to_same_var();
+            if let Some(insertable_for_block) = inseratables.get(&block.tid) {
+                block.propagate_input_expressions(Some(insertable_for_block.clone()));
+            }
+        }
+    }
```
Enkelmann, line 210: I am wary of potential undesired interactions of the `merge_def_assignments_to_same_var()` function and the computation of the input expression propagation. The most prudent way would be to use `merge_def_assignments_to_same_var()` before even computing the fixpoint for the expression propagation. This way we could avoid any unexpected interactions that might occur if we do it in another order.

Enkelmann (changes requested):
There are still some bugs in the implementation. But everything I found in the review should be easy to fix. :)

I have not yet reviewed the unit tests. Remind me to review them in my next review!

src/cwe_checker_lib/src/analysis/expression_propagation/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+/// The context struct for the expression propagation fixpoint computation.
+///
+/// The computation is a intra procedural forward fixpoint calculation
+/// that stores at each node the set of registers with their propagated expressions.
+/// This expressions can be used for expression specification among basic blocks.
+pub struct Context<'a> {
+    graph: &'a Graph<'a>,
+
+    /// Variable-expression lookup table primitive.
+    pub insertable_expressions: HashMap<Variable, Expression>,
```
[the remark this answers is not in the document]
m-rtz: Good point! On the downside Clippy annotates: "field `insertable_expressions` is never read".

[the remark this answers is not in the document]
Enkelmann: Yes, because you do not use this field in the Context struct. So why didn't you remove it?

Enkelmann (changes requested):

src/cwe_checker_lib/src/analysis/expression_propagation/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+                    if let Some(expr) = insertable_expressions.get(input_var) {
+                        extended_expression.substitute_input_var(input_var, expr)
+                    }
+                }
+                insertable_expressions.insert(var.clone(), extended_expression.clone());
+
+                // Expressions dependent on the assigned variable are no longer insertable.
+                // Except, if the assigned variable depends on itself.
+                insertable_expressions.retain(|input_var, input_expr| {
+                    !(input_var != var && input_expr.input_vars().into_iter().any(|x| x == var))
```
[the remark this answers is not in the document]
Enkelmann: Since the acceptance tests still fail after you resolved the other bugs, it is safe to assume that this code here is indeed incorrect. So we have to remove all `(var, expr)` pairs from the insertable expressions where the `expr` depends on the newly assigned variable. Without exceptions to this rule.

src/cwe_checker_lib/src/analysis/expression_propagation/mod.rs
```diff
@@ -0,0 +1,229 @@
+//! This module contains a fixpoint computation for intra-functional expression propagation of variables
```
Enkelmann, line 1: Some nitpicking: `intra-procedural` instead of `intra-functional`. Also, I would remove `of variables` from the sentence.

src/cwe_checker_lib/src/analysis/expression_propagation/mod.rs
```diff
@@ -0,0 +1,229 @@
...
+/// The computation is a intra procedural forward fixpoint calculation
+/// that stores at each node the set of registers with their propagated expressions.
+/// This expressions can be used for expression propagation among basic blocks.
+pub struct Context<'a> {
+    graph: &'a Graph<'a>,
+}
+
+impl<'a> Context<'a> {
+    /// Create a new context object for the given project and control flow graph.
+    pub fn new(_project: &'a Project, graph: &'a Graph) -> Context<'a> {
```
Enkelmann, line 24: If you do not need the `project` parameter, then you should remove it.

src/cwe_checker_lib/src/analysis/expression_propagation/mod.rs
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    fn update_call(
+        &self,
+        _value: &Self::Value,
+        _call: &Term<Jmp>,
+        _target: &Node,
+        _calling_convention: &Option<String>,
+    ) -> Option<Self::Value> {
+        None // This propagation is intra-functional
```
Enkelmann, line 108: `intra-procedural` ;-)

Enkelmann (changes requested):
Looks good! :-) Only some minor typos to fix before it can be merged.

Enkelmann (approved):

PR #360, opened 2022-11-14
Worklist order edit

m-rtz:
Introduces a new worklist order for the forward interprocedural fixpoint computation. This order prefers nodes of callees and is verified by a test.

### DIFF
--- a/src/cwe_checker_lib/src/analysis/forward_interprocedural_fixpoint.rs
+++ b/src/cwe_checker_lib/src/analysis/forward_interprocedural_fixpoint.rs
@@ -296,7 +296,7 @@ pub fn create_computation_with_bottom_up_worklist_order<'a, T: Context<'a>>(
     problem: T,
     default_value: Option<T::Value>,
 ) -> super::fixpoint::Computation<GeneralizedContext<'a, T>> {
-    let priority_sorted_nodes: Vec<NodeIndex> = create_bottom_up_worklist(&problem.get_graph());
+    let priority_sorted_nodes: Vec<NodeIndex> = create_bottom_up_worklist(problem.get_graph());
     let generalized_problem = GeneralizedContext::new(problem);
     super::fixpoint::Computation::from_node_priority_list(
         generalized_problem,
--- a/src/cwe_checker_lib/src/analysis/forward_interprocedural_fixpoint.rs
+++ b/src/cwe_checker_lib/src/analysis/forward_interprocedural_fixpoint.rs
@@ -268,22 +268,6 @@ pub fn create_computation<'a, T: Context<'a>>(
     super::fixpoint::Computation::new(generalized_problem, default_value.map(NodeValue::Value))
 }
 
-/// Generate a new computation from the corresponding context and an optional default value for nodes.
-/// Uses the alternate worklist order when computing the fixpoint.
-///
-/// The alternate worklist order moves nodes with 10 or more incoming edges to the end of the priority queue.
-/// This can improve the convergence speed for these nodes in the fixpoint algorithm.
-/// Use if you encounter convergence problems for nodes with a lot of incoming edges.
-pub fn create_computation_with_alternate_worklist_order<'a, T: Context<'a>>(
-    problem: T,
-    default_value: Option<T::Value>,
-) -> super::fixpoint::Computation<GeneralizedContext<'a, T>> {
-    let generalized_problem = GeneralizedContext::new(problem);
-    super::fixpoint::Computation::new_with_alternate_worklist_order(
-        generalized_problem,
-        default_value.map(NodeValue::Value),
-    )
-}
 /// Returns a node ordering with callee nodes behind caller nodes.
 pub fn create_bottom_up_worklist(graph: &Graph) -> Vec<NodeIndex> {
     let mut graph = graph.clone();
--- a/src/cwe_checker_lib/src/analysis/forward_interprocedural_fixpoint.rs
+++ b/src/cwe_checker_lib/src/analysis/forward_interprocedural_fixpoint.rs
@@ -284,21 +284,35 @@ pub fn create_computation_with_alternate_worklist_order<'a, T: Context<'a>>(
         default_value.map(NodeValue::Value),
     )
 }
+/// Returns a node ordering with callee nodes behind caller nodes.
+pub fn create_bottom_up_worklist(graph: &Graph) -> Vec<NodeIndex> {
+    let mut graph = graph.clone();
+    graph.retain_edges(|frozen, edge| !matches!(frozen[edge], Edge::Call(..)));
+    petgraph::algo::kosaraju_scc(&graph)
+        .into_iter()
+        .flatten()
+        .collect()
+}
+
+/// Returns a node ordering with caller nodes behind callee nodes.
+pub fn create_top_down_worklist(graph: &Graph) -> Vec<NodeIndex> {
+    let mut graph = graph.clone();
+    graph.retain_edges(|frozen, edge| !matches!(frozen[edge], Edge::CrReturnStub));
+    petgraph::algo::kosaraju_scc(&graph)
+        .into_iter()
+        .flatten()
+        .collect()
+}
 
 /// Generate a new computation from the corresponding context and an optional default value for nodes.
-/// Uses the Callee before Caller worklist order when computing the fixpoint.
-pub fn create_computation_with_callee_before_caller_worklist_order<'a, T: Context<'a>>(
+/// Uses a bottom up worklist order when computing the fixpoint.
+///
+/// The worklist order prefers callee nodes before caller nodes.
+pub fn create_computation_with_bottom_up_worklist_order<'a, T: Context<'a>>(
     problem: T,
     default_value: Option<T::Value>,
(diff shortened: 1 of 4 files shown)


src/cwe_checker_lib/src/analysis/forward_interprocedural_fixpoint.rs
```diff
@@ -283,3 +284,149 @@ pub fn create_computation_with_alternate_worklist_order<'a, T: Context<'a>>(
...
+/// Uses the Callee before Caller worklist order when computing the fixpoint.
+pub fn create_computation_with_callee_before_caller_worklist_order<'a, T: Context<'a>>(
+    problem: T,
+    default_value: Option<T::Value>,
+) -> super::fixpoint::Computation<GeneralizedContext<'a, T>> {
+    let mut graph = problem.get_graph().clone();
+    graph.retain_edges(|frozen, edge| !matches!(frozen[edge], Edge::Call(..)));
+    let priority_sorted_nodes: Vec<NodeIndex> = petgraph::algo::kosaraju_scc(&graph)
+        .into_iter()
+        .rev()
```
Enkelmann, line 298: This `rev()` confuses me. We did not reverse the order in the old node priority lists. Was this wrong?

Edit: Oh, maybe I explained the order logic the wrong way around. Nodes with a higher index in the priority list (i.e. towards the end of the list) have a higher priority in the worklist algorithm (i.e. get stabilized sooner). So I think this `rev()` is wrong here.

src/cwe_checker_lib/src/analysis/forward_interprocedural_fixpoint.rs
```diff
@@ -283,3 +284,149 @@ pub fn create_computation_with_alternate_worklist_order<'a, T: Context<'a>>(
         default_value.map(NodeValue::Value),
     )
 }
+
+/// Generate a new computation from the corresponding context and an optional default value for nodes.
+/// Uses the Callee before Caller worklist order when computing the fixpoint.
+pub fn create_computation_with_callee_before_caller_worklist_order<'a, T: Context<'a>>(
```
Enkelmann, line 290: `bottom_up_worklist_order` is in my opinion a better name than `callee_before_caller_worklist_order`, since bottom-up analysis is a common term for this.

src/cwe_checker_lib/src/analysis/forward_interprocedural_fixpoint.rs
```diff
@@ -283,3 +284,149 @@ pub fn create_computation_with_alternate_worklist_order<'a, T: Context<'a>>(
...
+        .collect();
+
+    let generalized_problem = GeneralizedContext::new(problem);
+    super::fixpoint::Computation::from_node_priority_list(
+        generalized_problem,
+        default_value.map(NodeValue::Value),
+        priority_sorted_nodes,
+    )
+}
+#[cfg(test)]
```
Enkelmann, line 309: Can you leave an empty line before `#[cfg(test)]`? I am surprised that `cargo fmt` apparently has no opinion about this...

src/cwe_checker_lib/src/analysis/forward_interprocedural_fixpoint.rs
```diff
@@ -283,3 +284,149 @@ pub fn create_computation_with_alternate_worklist_order<'a, T: Context<'a>>(
...
+            Some(HashMap::new()),
+        );
+        // The first two nodes should belong to the callee
+        for node in comp.get_worklist()[0..1].iter() {
+            match graph[*node] {
+                crate::analysis::graph::Node::BlkStart(_, sub)
+                | crate::analysis::graph::Node::BlkEnd(_, sub) => {
+                    assert_eq!(
+                        sub.term,
+                        project.program.term.subs.get(&sub.tid).unwrap().term
```
Enkelmann, line 425: Why not compare TIDs here instead? I know, it is just a unit test, so speed is not important, but it still feels unnecessary to compare the whole functions if comparing their TIDs would  be enough.

Enkelmann (changes requested):
If I remember correctly, we wanted to implement a top-down worklist order in addition to the bottom-up worklist order.

Also, why did you implement this in `forward_interprocedural_fixpoint.rs` instead of `fixpoint.rs`?

In `fixpoint.rs` the old `new_with_alternate_worklist_order` method should be replaced by your new function, since the alternate worklist order is deprecated and all analyses using it are now bottom-up analyses, if I remember correctly.

src/cwe_checker_lib/src/analysis/forward_interprocedural_fixpoint.rs
```diff
@@ -267,19 +268,207 @@ pub fn create_computation<'a, T: Context<'a>>(
...
+            },
+        };
+
+        let caller_block_2 = Term {
+            tid: Tid::new("caller_block_2"),
+            term: Blk {
+                defs: vec![],
+                jmps: vec![],
+                indirect_jmp_targets: Vec::new(),
+            },
```
Enkelmann, line 365: You could write a helper function `new_block(name: &str) -> Term<Blk>` that creates such an empty block. If you replace all three block creations in this function with such a helper function you could save one or 2 lines.

src/cwe_checker_lib/src/analysis/forward_interprocedural_fixpoint.rs
```diff
@@ -267,19 +268,207 @@ pub fn create_computation<'a, T: Context<'a>>(
...
+            .program
+            .term
+            .subs
+            .insert(caller_function.tid.clone(), caller_function.clone());
+
+        project
+            .program
+            .term
+            .subs
+            .insert(called_function.tid.clone(), called_function.clone());
```
Enkelmann, line 411: a shorter way to write this:
```rust
project.program.term.subs = BTreeMap::from([
    (caller_function.tid.clone(), caller_function.clone()),
    (called_function.tid.clone(), called_function.clone()),
]);

File: src/cwe_checker_lib/src/analysis/forward_interprocedural_fixpoint.rs [hunk 7]
```diff
@@ -267,19 +268,207 @@ pub fn create_computation<'a, T: Context<'a>>(
...
+    /// Checks if the nodes corresponding to the callee function are first in the worklist.
+    fn check_bottom_up_worklist() {
+        let project = mock_project();
+        let extern_subs = project
+            .program
+            .term
+            .extern_symbols
+            .keys()
+            .cloned()
+            .collect();
```
Enkelmann, line 426: Since you know this list to be empty, you could also just create an empty list for your test.

src/cwe_checker_lib/src/analysis/forward_interprocedural_fixpoint.rs
```diff
@@ -267,19 +268,207 @@ pub fn create_computation<'a, T: Context<'a>>(
...
+        let comp = create_computation_with_bottom_up_worklist_order(context, Some(HashMap::new()));
+        // The last two nodes should belong to the callee
+        for node in comp.get_worklist()[6..].iter() {
+            match graph[*node] {
+                crate::analysis::graph::Node::BlkStart(_, sub)
+                | crate::analysis::graph::Node::BlkEnd(_, sub) => {
+                    assert_eq!(
+                        sub.tid,
+                        project.program.term.subs.get(&sub.tid).unwrap().tid
+                    )
```
Enkelmann, line 439: This is an empty test, since you basically ask `x == x` here. Replace `project.program.term.subs.get(&sub.tid).unwrap().tid` with `Tid::new("called_function")` to fix this.

src/cwe_checker_lib/src/analysis/forward_interprocedural_fixpoint.rs
```diff
@@ -267,19 +268,207 @@ pub fn create_computation<'a, T: Context<'a>>(
...
+        let comp = create_computation_with_bottom_up_worklist_order(context, Some(HashMap::new()));
+        // The first two nodes should belong to the callee
+        for node in comp.get_worklist()[..2].iter() {
+            match graph[*node] {
+                crate::analysis::graph::Node::BlkStart(_, sub)
+                | crate::analysis::graph::Node::BlkEnd(_, sub) => {
+                    assert_eq!(
+                        sub.tid,
+                        project.program.term.subs.get(&sub.tid).unwrap().tid
+                    )
```
Enkelmann, line 468: As above, replace `project.program.term.subs.get(&sub.tid).unwrap().tid` with `Tid::new("called_function")` to fix the test.

Enkelmann (changes requested):
The code looks good and is almost ready to be merged. :-)
- I found one little bug in your unit test code that should be fixed. The other code comments are optional clean code hints.
- Can you remove the old function that constructs the "alternate_worklist_order" in your PR? Since you implement the replacement functions here, it makes sense to remove this now-deprecated function in the same PR.

src/cwe_checker_lib/src/analysis/forward_interprocedural_fixpoint.rs
m-rtz: Oh yes, I see. Thank you!

Enkelmann (approved):

PR #364, opened 2022-12-10
Test refactoring step 1

m-rtz:
First changes for introducing macros to the test code base. All further changes are done as PR to macros_and_test_changes for small sized partial reviews.

### DIFF
--- a/src/cwe_checker_lib/src/intermediate_representation/macros.rs
+++ b/src/cwe_checker_lib/src/intermediate_representation/macros.rs
@@ -12,18 +12,17 @@
 /// ```
 #[macro_export]
 macro_rules! variable {
-    (  $x:literal  ) => {{
+    (  $x:expr  ) => {{
         parsing::parse_variable($x)
     }};
 }
 
-/// Creates a `Bitvector` specified by the string slice of form `0xvalue:size`.
+/// Creates a `Bitvector` specified by the string slice of form `0xvalue:size` or value:size.
 ///
-/// `value` is in hexadecimal representation.
+/// `value` is either in hexadecimal representation with leading `0x` or in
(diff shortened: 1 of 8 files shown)


src/cwe_checker_lib/src/intermediate_representation/macros.rs
```diff
@@ -0,0 +1,549 @@
...
+#[macro_export]
+macro_rules! def {
+    ($x:expr) => {
+        parsing::parse_def($x, 0)
+    };
+}
+
+pub mod parsing {
+    //! Provides parsing functions for the macros defined in `macros.rs`.
+    //! This module hides the parsing functions and allows exposure of the macros only.
```
Enkelmann, line 138: To completely hide the parsing functions you should make the module private and add a `use parsing::*;` afterwards. This way the module would be completely inaccessible outside of the `macros` module.

Enkelmann (commented):
We should consider changing behavior of the expression printing functions so that the printed sizes are in bytes and not in bits. This would make it consistent with the macros. However, this would mean that we should look at all print and `to_json_compact` functions to ensure that we do it consistently everywhere. So this is maybe something for the next PR.

src/cwe_checker_lib/src/intermediate_representation/macros.rs
m-rtz: I am getting the `error failed to resolve: use of undeclared crate or module parsing
use of undeclared crate or module parsing` originating from the analysis' tests. This might be due to the fact, that the module `parsing` and the contained function is not in scope there but its call is "injected" by the macro.
Maybe leaving the module as `pub` is a more or less clean way to make the function usable everywhere? It will not be hidden, but maybe we can not do this at all?

PR #370, opened 2023-01-12
fixed bug of stack_alignment_substitution analysis

m-rtz:
The stack_alignment_substitution-Analysis only considered the first block of every sub for replacing AND-Operations. If a sub has leading blocks without defs, this analysis does not alter the function at all.
Such empty blocks might be generated by `endbr`-like instructions, which sometimes are the first instructions of a function.

The analysis was edited and now skips leading, empty blocks of a function.

### DIFF
--- a/src/cwe_checker_lib/src/analysis/stack_alignment_substitution/mod.rs
+++ b/src/cwe_checker_lib/src/analysis/stack_alignment_substitution/mod.rs
@@ -101,67 +101,70 @@ pub fn substitute_and_on_stackpointer(project: &mut Project) -> Option<Vec<LogMe
 
     'sub_loop: for sub in project.program.term.subs.values_mut() {
         let journaled_sp: &mut i64 = &mut 0;
-        // only for the first block SP can be reasonable tracked
-        if let Some(blk) = sub.term.blocks.first_mut() {
-            for def in blk.term.defs.iter_mut() {
-                if let Def::Assign { var, value } = &mut def.term {
-                    if *var == project.stack_pointer_register {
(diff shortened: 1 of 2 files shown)


Enkelmann (commented):
Hmm... this is a somewhat hacky solution. We do not guarantee anything about the order of blocks in a sub except for the first block. So to ensure that the second block is really the second block one would have to follow the jump at the end of the first block instead of taking the second block in the `blocks` array.

Nevertheless, i would be willing to merge this for the time being. In most cases it should still work and I think it is unlikely to introduce random errors by substituting the wrong instructions. However, if we use the hacky solution, could you add a code comment that this is indeed *hacky*?

src/cwe_checker_lib/src/analysis/stack_alignment_substitution/mod.rs
```diff
@@ -86,6 +88,59 @@ fn journal_sp_value(
     }
 }
 
+/// Returns the tid of the target of the first Jmp::Branch of the provided block.
+fn get_first_branch_tid(blk: &Term<Blk>) -> Option<&Tid> {
+    for jmp in &blk.term.jmps {
```
Enkelmann, line 93: I think this should be
```rust
if let Some(jmp) = blk.term.jmps.get(0) {
```
The `for`-loop would otherwise also catch one branch of conditional jumps, which is not what we want here.

Enkelmann (changes requested):
One small issue, but the rest looks good to me.

Enkelmann (approved):

PR #381, opened 2023-01-30
Test refactoring

m-rtz:
Another intermediate step of refactoring test code base. Next are the remaining modules of `intermediate_representation`.


src/cwe_checker_lib/src/analysis/expression_propagation/mod.rs
```diff
@@ -191,7 +191,7 @@ fn insert_expressions(
 ) {
     for sub in program.subs.values_mut() {
         for block in sub.term.blocks.iter_mut() {
-            block.propagate_input_expressions(inseratables.get(&block.tid).cloned());
+            propagate_input_expressions(block, inseratables.get(&block.tid).cloned());
```
Enkelmann, line 194: Could you also fix the typo `inseratables`instead of `insertables` here? I know, it has nothing to do with your PR, but we cannot just leave it there...

Enkelmann (commented):
Other than the typo (which was in the code prior to the PR) everything looks good to me.

Enkelmann (approved):

PR #383, opened 2023-02-02
Test refactoring

m-rtz:
Allowed `_` as part of variable names, refactored tests


src/cwe_checker_lib/src/intermediate_representation/term/builder.rs
```diff
@@ -58,3 +90,507 @@ impl Jmp {
...
+            },
+        }
+    }
+    /// Creates empty block with tid "block".
+    pub fn mock() -> Term<Blk> {
+        Self::mock_with_tid("block")
+    }
+}
+#[cfg(test)]
+impl Sub {
```
Enkelmann, line 159: We should add an empty line after the closing bracket of the previous item.

src/cwe_checker_lib/src/intermediate_representation/term/builder.rs
```diff
@@ -58,3 +90,507 @@ impl Jmp {
...
+            term: Sub {
+                name: name.to_string(),
+                blocks: Vec::new(),
+                calling_convention: None,
+            },
+        }
+    }
+}
+/// Wrapper for subpiece to model float register for argument passing
+#[cfg(test)]
```
Enkelmann, line 172: We should add an empty line after the closing bracket of the previous item.

src/cwe_checker_lib/src/intermediate_representation/term/builder.rs
```diff
@@ -58,3 +90,507 @@ impl Jmp {
...
+    size: u64,
+) -> Expression {
+    Expression::subpiece(
+        expr!(format!("{name}:{reg_size}")),
+        ByteSize::new(low_byte),
+        ByteSize::new(size),
+    )
+}
+#[cfg(test)]
+impl CallingConvention {
```
Enkelmann, line 186: We should add an empty line after the closing bracket of the previous item.

Enkelmann (changes requested):
While I like the decision to bundle all the builder-like helper function in `builder.rs`, the resulting file is too long. We should split it up in at least 2 files.

Also, could you look through `builder.rs` and add missing empty lines between items? There were more cases than the three I explicitly flagged.

Enkelmann (approved):
Looks good to me now. The CI failure seems to be unrelated, so I am going to merge this anyway.

Edit: Oh, I missed the documentation errors! Can you fix them please?

PR #387, opened 2023-02-07
Typo: we try to treat 'is' as ....

Brubbish:
That's it.


Enkelmann (approved):
Thank you for the contribution! Merging.

PR #389, opened 2023-02-16
Function sig param sanitization

m-rtz:


src/cwe_checker_lib/src/analysis/function_signature/mod.rs
```diff
@@ -266,6 +278,111 @@ impl FunctionSignature {
...
+        let first_arg_offset = first_arg.eval_stack_offset()?.try_to_u64()?;
+        let first_arg_size = u64::from(*first_size);
+        let second_arg_offset = second_arg.eval_stack_offset()?.try_to_u64()?;
+        let second_arg_size = u64::from(*second_size);
+
+        if first_datatype == second_datatype {
+            let mut logs = vec![];
+            let first_arg_upper_bound = first_arg_offset + first_arg_size;
+            // Check if they intersect
+            if first_arg_upper_bound >= second_arg_offset {
```
Enkelmann, line 365: I think this comparison needs to be `>`, not `>=`. We should add a corresponding unit test.

src/cwe_checker_lib/src/analysis/function_signature/mod.rs
```diff
@@ -266,6 +278,111 @@ impl FunctionSignature {
...
+                    // second arg is not a subset, but intersects
+                    logs.push(
+                        "Merged a stack parameter, that intersect another but is no subset"
+                            .to_string(),
+                    );
+                }
+                let merged_size = second_arg_upper_bound - first_arg_offset;
+                return Ok(((first_address.clone(), merged_size), logs));
+            } else {
+                return Err(anyhow!("Args do not intersect"));
```
Enkelmann, line 379: You do not actually use the error cases, do you? If not, you could change the return type of the function to an `Option` instead of a `Result`.

src/cwe_checker_lib/src/analysis/function_signature/mod.rs
```diff
@@ -305,4 +422,40 @@ pub mod tests {
...
         }
     }
+
+    #[test]
+    fn test_parameter_merging() {
+        let mut func_sig = FunctionSignature::mock_x64();
+        let stack_parm_1 = Arg::Stack {
+            address: expr!("RSP:8 + 0x1000:8"),
+            size: 8.into(),
+            data_type: Some(Datatype::Integer),
```
Enkelmann, line 432: This is just nitpicking, but I think we should leave the data types as `None` even in the unit tests. It just should never occur that we already have some data type information in the function signature analysis.

Enkelmann (changes requested):
As mentioned in the other comments, we need some more unit tests to cover more cases that are probably not handled correctly right now. Also, the unit test should probably test the whole `sanitize` function, so that we can also catch unwanted interactions with future additions to the `sanitize` function with the tests.

src/cwe_checker_lib/src/analysis/function_signature/mod.rs
```diff
@@ -257,97 +256,109 @@ impl FunctionSignature {
...
+                }
             }
         }
-        for arg in removed_args {
-            self.parameters.remove(&arg);
-        }
-        self.parameters.extend(merged_args);
-        Ok(logs)
+
+        //self.parameters = stack_parms.into_iter().collect();
```
Enkelmann, line 336: Did you forget to remove this commented-out line?

Enkelmann (changes requested):
Almost ready to merge. :-)

PR #407, opened 2023-04-27
Pcode extractor rewrite

m-rtz:
First of a few changes for rewriting the pcode extraction procedure.
This changes affect the Java side and pcode is now passed to the Rust code base without modification nor pre-processing.


src/ghidra/p_code_extractor/PcodeExtractor.java
```diff
@@ -1,288 +1,249 @@
...
+            writer.close();
+        } catch (JsonIOException e) {
+            e.printStackTrace();
+        } catch (IOException e) {
+            e.printStackTrace();
         }
-    }
 
+        println("Pcode was successfully extracted!");
+    }
```
Enkelmann, line 126: If an error is catched here, then we should not print the "success" message. Also: Are there string error messages that we can print together with the stack traces here?

src/ghidra/p_code_extractor/PcodeExtractor.java
```diff
@@ -1,288 +1,249 @@
...
-     * Checks whether the assembly instruction is a nop instruction and adds a jump to the block.
-     * Checks whether a jump occured within a ghidra generated pcode block and fixes the control flow
-     * by adding missing jumps between artificially generated blocks.
-     * Checks whether an instruction is in a delay slot and, if so, ignores it 
-     * as Ghidra already includes the instruction before the jump
+     * This functions returns the base register, e.g. RDI is returned and not EDI.
+     * *Approximation*: This function returns the first 10 or less integer registers
+     * This is due to the only way of extracting the integer parameter register
+     * available
+     * (https://ghidra.re/ghidra_docs/api/ghidra/program/model/lang/PrototypeModel.html)
```
Enkelmann, line 136: We definitely have to revisit this when the Rust-side of the parsing is implemented. Just to make sure whether we want to fall back to the old "directly parse the xml file" behaviour or not.

src/ghidra/p_code_extractor/VarnodeSimple.java
```diff
@@ -0,0 +1,42 @@
...
+        this.addressspace = varnode.getAddress().getAddressSpace().getName();
+        this.id = varnode.getAddress().toString("0x");
+        if (context.getRegister(varnode) != null) {
+            this.id = context.getRegister(varnode).getName();
+        }
+    }
+
+    public VarnodeSimple(Register register) {
+        this.id = register.getName();
+        this.size = (int) register.getBitLength() / 8;
```
Enkelmann, line 30: If I remember correctly, this code should work as intended. Especially if this is just copied from the old code, then it should be correct. But just to make sure: Have you tested it with flag registers?

Enkelmann (changes requested):

src/ghidra/p_code_extractor/VarnodeSimple.java
m-rtz: It originates from the old code. Flag registers get a (byte) size of 1.

src/ghidra/p_code_extractor/PcodeExtractor.java
m-rtz: I will keep that in mind. Nevertheless it should come up during parsing the pcode.

src/ghidra/p_code_extractor/PcodeExtractor.java
m-rtz: Good catch!

Enkelmann (approved):
LGTM

PR #413, opened 2023-06-12
Pcode parser rewrite

m-rtz:
An intermediate step of rewriting the pcode parser. Still in development and contains debug/testing constructions.


Enkelmann (changes requested):
The `ghidra_pcode/mod.rs` is already over 400 lines long. We should think about splitting it up in multiple files or we will end up with an extremely long file (like the one in the old pcode-parsing module).

src/cwe_checker_lib/src/ghidra_pcode/mod.rs
```diff
@@ -0,0 +1,414 @@
...
+    /// virtual registers.
+    ///
+    /// Returns `Err` if the addressspace is neither `"const"`, `"register"` nor `"unique"`.
+    fn into_ir_expr(self) -> Result<Expression> {
+        println!("\t{} : {} : {}", self.addressspace, self.id, self.size);
+        match self.addressspace.as_str() {
+            "const" => Ok(Expression::Const(Bitvector::from_u64(u64::from_str_radix(
+                self.id.trim_start_matches("0x"),
+                16,
+            )?))),
```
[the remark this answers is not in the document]
m-rtz: Good catch!

src/cwe_checker_lib/src/ghidra_pcode/mod.rs
```diff
@@ -0,0 +1,414 @@
...
+            let mut explicit_loads = self.create_implicit_loads(address);
+            defs.append(&mut explicit_loads);
+        }
+        if self.has_implicit_store() {
+            todo!()
+        }
+
+        let def = match self.pcode_mnemonic {
+            PcodeOperation::ExpressionType(expr_type) => self.create_def(address, expr_type),
+            PcodeOperation::JmpType(jmp_type) => todo!(),
```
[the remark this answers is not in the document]
m-rtz: This is not yet implemented. In future, if a pcode JmpType occures it is supposed to be translated into its IR representative. I have not have a deeper look into this yet, but in the best case the user does not need to interact.

src/cwe_checker_lib/src/ghidra_pcode/tests.rs
```diff
@@ -5,43 +5,51 @@ use crate::{bitvec, def, expr, variable};
...
+            pcode_index: self.pcode_index,
+            pcode_mnemonic: mnemonic,
+            input0: self.input0.clone(),
+            input1: self.input1.clone(),
+            input2: self.input2.clone(),
+            output: self.output.clone(),
+        }
+    }
+    fn set_varnodes(
+        mut self,
```
Enkelmann, line 43: These two functions should again use the `with_x` naming scheme instead of the `set_x` naming scheme, since they use the builder pattern where one need to use the returned value after calling the function.

src/cwe_checker_lib/src/intermediate_representation/term/builder_low_lvl.rs
```diff
@@ -1,5 +1,6 @@
 //! This module contains the implementations of various builder functions
-//! for lower intermediate representation terms [Expression](crate::intermediate_representation::Expression),
+//! for lower intermediate representation terms [Term](crate::intermediate_representation::Term)
```
Enkelmann, line 2: Typo: There is a comma missing at the end of the line.

src/cwe_checker_lib/src/intermediate_representation/term/builder_low_lvl.rs
```diff
@@ -11,6 +12,16 @@
 #[cfg(test)]
 use crate::{expr, intermediate_representation::*, variable};
 
+#[cfg(test)]
+impl<T> Term<T> {
+    /// Sets the field `id` of `Tid`
+    #[cfg(test)]
+    pub fn with_tid_id(mut self, id: String) -> Term<T> {
```
Enkelmann, line 19: to nitpick a little: If the whole `impl`-block is `#[cfg(test)]`, you do not need to add `#[cfg(test)]` to each of the items.

Enkelmann (changes requested):

Enkelmann (approved):
LGTM, merging.

PR #439, opened 2024-02-14
Check for CWE-337

PascalBeyer:
Implements a check for CWE-337: Predictable Seed in Pseudo-Random Number Generator (PRNG)

The check works by reusing the taint-tracking code from the check for CWE-467 to track return values of "source" functions (e.g.: `time`) to arguments of "sink" functions (e.g: `srand`).


test/src/lib.rs
```diff
@@ -394,6 +394,31 @@ mod tests {
...
+    fn cwe_337() {
+        let mut error_log = Vec::new();
+        let mut tests = all_test_cases("cwe_337", "CWE337");
+
+        mark_architecture_skipped(&mut tests, "ppc64"); // Ghidra generates mangled function names here for some reason.
+        mark_architecture_skipped(&mut tests, "ppc64le"); // Ghidra generates mangled function names here for some reason.
+
+        mark_skipped(&mut tests, "x86", "gcc"); // x86 uses the stack for return values/arguments, the check is only register based.
+        mark_skipped(&mut tests, "x86", "clang"); // x86 uses the stack for return values/arguments, the check is only register based.
+        mark_compiler_skipped(&mut tests, "mingw32-gcc"); // x86 uses the stack for return values/arguments, the check is only register based.
```
Enkelmann, line 408: Your line would also skip the x64-mingw32-gcc test case, which we do not want to, because it works just fine (since it is not 32-bit x86). Actually, you can combine the three x86-specific lines to just
```rust
mark_architecture_skipped(&mut tests, "x86");
```

Enkelmann (changes requested):
Only two minor things to fix remaining before we can merge the check.

PR #443, opened 2024-02-28
Bump minimal supported Rust version to 1.73

vobst:
Also add minimal supported Rust version to `Cargo.toml` files.

Add the minimal supported Rust version to `Cargo.toml` files to fail the
build as early as possible if the used Rust version is too old.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM rust:1.73 AS builder
+FROM rust:1.73-bullseye AS builder
 
 WORKDIR /cwe_checker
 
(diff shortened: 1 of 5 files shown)


Enkelmann:
As I feared, we need to update the `ghidra_headless_base` image, because its contained glibC is too old. I will talk to @jstucke about it.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM rust:1.70 AS builder
+FROM rust:1.73 AS builder
```
Enkelmann, line 1: In the meantime I found a better long-term solution: Can you replace `rust:1.73` with `rust:1.73-bullseye` ? That pins the underlying Debian version for the builder image to one that works with our `ghidra_headless_base` image.

Enkelmann (approved):

PR #451, opened 2024-03-18
Check for CWE 252

vobst:
This PR tracks the progress on the addition of a check for CWE252. It is based on the Taint Analysis abstractions introduced by PR #450. While in draft state, it will be rebased and force-pushed without notice!

Use the new Taint Analysis abstraction to implement a check for CWE252. The check performs a limited interprocedural analysis.

# TODO

- [x] make it compile
- [x] propagate memory objects
- [x] settle on taint model
  - [x] one taint and many passes vs. many taints and one pass (Decision: one taint and many passes. Reason: generic taint is something for its own PR)
- [x] build initial list of "must check" functions:
  - [x] libc
  - [x] Linux
- [ ] tune propagation/warning behavior to balance FP vs FN
- [ ] add unit tests
- [x] add integration/acceptance tests

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,8 @@
 0.9-dev
 ===
 
+-   Improve taint analysis abstraction to simplify interprocedural bottom-up
+    analysis of memory taint (PR #451)
 -   Added check for CWE-252: Unchecked Return Value (PR #451)
 -   Implemented experimental support for Linux Kernel Modules (PR #441)
 -   Support LZCOUNT P-Code operation (PR #444)


Enkelmann:
In the short term, the huge list of symbol names in `lkm_config.json` is probably OK. In the long term, we still need to find a better solution, especially since the list may go out of sync with newer Linux kernel versions over time. Maybe provide or document a way how to extract the function list from Linux kernel sources? On the other hand, that would be more inconvenient for users, because they would need to do it before running the cwe_checker...

Enkelmann (changes requested):
The code looks good, I only found minor issues.

But we still need tests! Most importantly, we need an acceptance test.

Enkelmann:
> I am also not super happy with the huge list in the repo, maybe we can start with a trimmed down list, e.g., only API functions that are defined by the core kernel (~4k).

I think that would be a good compromise. 4k is at least a lot less than 15k.

> Currently the process I use to generate this list is not something I'd consider ready for publication. Maybe one can do something more elegant but that would be a bigger project.

Then let us postpone it for the time being and stick to using the resulting list in this PR. If it remains a bigger project, we could also think about moving it into a repository separate from the core cwe_checker code.

> The list getting out-of-sync isn't that much of a problem though - in the end we want to be able to analyze modules that were compiled for all sorts of kernel versions. Thus, I guess in the long term we should periodically regenerate the list, but take the union with the previous one.

Good point!

src/cwe_checker_lib/src/analysis/taint/state.rs
```diff
@@ -459,6 +449,154 @@ impl State {
...
+// FIXME: The used algorithm for merging the taints contained in memory
+// regions is unsound when merging taints that intersect only partially.
+// If, for example, in state A `RSP[0:3]` is tainted and in state B
+// `RSP[2:3]` is tainted, A u B will only report `RSP[2:3]` as tainted.
+//
+// For the NULL pointer dereference check, however, this should not have an
+// effect in practice, since these values are usually unsound or a sign of
+// incorrect behavior of the analysed program.
+// FIXME: This looks a lot like it should be a method on `MemRegion<T>`.
+fn merge_memory_object_with_offset(
```
Enkelmann, line 590: We should add a proper doc-comment to the function.

src/cwe_checker_lib/src/analysis/taint/state.rs
```diff
@@ -459,6 +449,154 @@ impl State {
...
+    mem_object: &mut MemRegion<Taint>,
+    other_memory_object: &MemRegion<Taint>,
+    offset: i64,
+) {
+    for (index, taint) in other_memory_object.iter() {
+        mem_object.insert_at_byte_index(*taint, *index + offset);
+        // FIXME: Unsound in theory for partially intersecting
+        // taints. Should not matter in practice.
+    }
```
Enkelmann, line 599: Should we also add a warning that this code will become wrong when we allow more complicated taint types in the future (because it does not use the merge function for taints)?

src/cwe_checker_lib/src/analysis/taint/state/tests.rs
```diff
@@ -58,6 +60,69 @@ fn new_pointer(location: &str, offset: i64) -> DataDomain<ValueDomain> {
...
+    let mut other_memory_object = MemRegion::<Taint>::new(ByteSize::new(8));
+
+    memory_object.insert_at_byte_index(taint_4, 2);
+    memory_object.insert_at_byte_index(taint_8, 8);
+    memory_object.insert_at_byte_index(taint_4, 16);
+    other_memory_object.insert_at_byte_index(taint_8, 0);
+    other_memory_object.insert_at_byte_index(taint_4, 8);
+    other_memory_object.insert_at_byte_index(taint_4, 14);
+
+    merge_memory_object_with_offset(&mut memory_object, &other_memory_object, 0);
```
Enkelmann, line 80: From my experience, it is important to also test the case when there is an actual offset other than zero. Just to be sure that offset computations are done without sign errors.

src/cwe_checker_lib/src/checkers/cwe_252/context.rs
```diff
@@ -0,0 +1,330 @@
...
+    project: &'a Project,
+    pi_result: &'a PointerInference<'b>,
+    /// Used to send generated CWE warnings to the collector.
+    cwe_sender: crossbeam_channel::Sender<CweWarning>,
+}
+
+impl<'a, 'b: 'a> TaCompCtx<'a, 'b> {
+    /// Creates a new taint analysis context for the given call to an external
+    /// function.
+    pub(super) fn new(
```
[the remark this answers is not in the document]
Enkelmann: But isn't the context module (and thus the context type) private to the `cwe_252` module anyway? In other words, `pub` and `pub(super)` should have the same effect. Which adds to my confusion, because explicitly using `pub(super)` here suggests otherwise.

src/cwe_checker_lib/src/checkers/cwe_252/context.rs
```diff
@@ -0,0 +1,330 @@
...
+            //   are thus not tainted.
+            // - There is a chance that caller saved registers will be restored
+            //   by the caller such that the taint is immediatly eliminated and
+            //   we catch cases where the called function has ignored tainted
+            //   values.
+            let mut new_state = state.clone();
+
+            new_state.remove_all_memory_taints();
+
+            Some(new_state)
```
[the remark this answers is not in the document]
Enkelmann: You are right, I forgot that. Actually, the cwe_checker will panic anyway during the function signature analysis if there is no standard calling convention to fall back to. But other code still handles this case without panicking. ... Ok, we will have to decide in a future PR if we want to always require a standard calling convention to exist (thus getting rid of these special cases) or not (which would require changes to the function signature analysis).

test/src/lib.rs
```diff
@@ -418,6 +424,44 @@ mod tests {
...
+        let mut error_log = Vec::new();
+        let mut tests = linux_test_cases("cwe_252", "CWE252");
+        let num_expected_occurences = 9;
+        let num_expected_occurences_lkm = 1;
+
+        tests.extend(lkm_test_cases("cwe_252", "CWE252"));
+
+        mark_architecture_skipped(&mut tests, "ppc64");
+        mark_architecture_skipped(&mut tests, "ppc64le");
+        mark_skipped(&mut tests, "aarch64", "gcc");
```
Enkelmann, line 439: Do you know the reason why it fails for aarch64-gcc? Or maybe I should ask why it fails for gcc instead... I am mainly interested in whether turning on optimizations has something to do with it or not.

test/src/lib.rs
```diff
@@ -418,6 +424,44 @@ mod tests {
...
+
+        for test_case in tests {
+            let num_expected_occurences = if test_case.is_lkm {
+                num_expected_occurences_lkm
+            } else {
+                num_expected_occurences
+            };
+
+            if let Err(error) = test_case.run_test("[CWE252]", num_expected_occurences) {
+                println!("{:?}", error);
```
Enkelmann, line 454: This `println!` should be unnecessary, because printing is done by `print_errors` later on.

Enkelmann (changes requested):

test/artificial_samples/SConstruct
```diff
@@ -3,7 +3,7 @@ import os
 build_path = 'build'
 
 supported_architectures = ['x64', 'x86', 'arm', 'aarch64', 'mips', 'mipsel', 'mips64', 'mips64el', 'ppc', 'ppc64', 'ppc64le']
-skip_for_pe = ['cwe_782.c', 'cwe_426.c', 'cwe_243.c', 'cwe_243_clean.c']
```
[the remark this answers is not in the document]
Enkelmann: If we use unoptimized code with `-O0` instead, we will not need `__attribute_no_inline__` at all. And then the tests should (in theory) also work with mingw.

src/cwe_checker_lib/src/abstract_domain/mod.rs
```diff
@@ -36,10 +36,32 @@ pub use domain_map::*;
 /// Each abstract domain is partially ordered.
 /// Abstract domains of the same type can be merged.
 pub trait AbstractDomain: Sized + Eq + Clone {
-    /// Return an upper bound (with respect to the partial order on the domain) for the two inputs `self` and `other`.
+    /// Computes the least upper bound (with respect to the partial order on the
+    /// domain) for the two inputs `self` and `other`.
```
Enkelmann: To nitpick a little: Merging is allowed to use widening when it deems it necessary. Which means that it does not guarantee to compute the least upper bound.

src/cwe_checker_lib/src/abstract_domain/mod.rs
```diff
@@ -36,10 +36,32 @@ pub use domain_map::*;
...
+    ///
+    /// Modifies `self` in-place to hold the result. This can be useful in
+    /// situations where it is not necessary to create a new object and more
+    /// efficient to modify an existing one in-place.
+    ///
+    /// # Default
+    ///
+    /// Calls [`AbstractDomain::merge`] on the inputs and overwrites `self` with
+    /// the result. Does nothing when `self` is equal to `other`.
+    fn merge_with(&mut self, other: &Self) {
```
Enkelmann: We could let this function return `&mut self` to support method chaining. I do not know whether we would use it that often, but it is nice to have just in case.

src/cwe_checker_lib/src/analysis/taint/state/memory_taint.rs
```diff
@@ -0,0 +1,212 @@
...
+                    merge_memory_object_with_offset(memory_object, other_memory_object, 0);
+                })
+                .or_insert_with(|| other_memory_object.clone());
+        }
+    }
+}
+
+/// Represents our knowledge about taint in memory at a particular point in the
+/// program.
+pub type MemoryTaint = DomainMap<AbstractIdentifier, MemRegion<Taint>, UnionMergeStrategy>;
```
Enkelmann: You do not actually use the `MemoryTaintMergeStrategy`? Is this an oversight here or is the `MemoryTaintMergeStrategy` dead code now?

I also do not see the difference to the `UnionMergeStrategy` at the moment.

test/lkm_samples/cwe_252.c
```diff
@@ -0,0 +1,40 @@
...
+
+    return 0;
+}
+
+static void __exit test_exit(void)
+{
+    pr_info("Goodbye, World\n");
+}
+
+MODULE_LICENSE("GPL v2");
```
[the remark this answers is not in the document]
Enkelmann: We can keep it. I just wondered whether we now have to mention in the `Readme.md` that parts of the repository are GPL instead of LGPL. But it is only test code, I guess we can just quietly ignore what we tell the Linux kernel here and treat it as licensed like the rest of the repository.

src/cwe_checker_lib/src/analysis/taint/state/memory_taint.rs
vobst: The point where it is used is in the `merge_with` method of `State`

```rust
    fn merge_with(&mut self, other: &Self) {
        self.register_taint.merge_with(&other.register_taint);
        self.memory_taint.merge_with(&other.memory_taint);
    }
```

which calls it via the `merge_with` implementation of `DomainMap`

```rust
    fn merge_with(&mut self, other: &Self) {
        if self == other {
            return;
        }

        let mut_map = Arc::make_mut(&mut self.inner);

        S::merge_map_with(mut_map, &other.inner);
    }
```

The difference to the `UnionMergeStrategy` is that it uses `merge_memory_object_with_offset` instead of the `merge` implementation of `MemRegion` (and this is in fact the only difference). This is to keep the current behavior of memory merging as-is.

test/artificial_samples/cwe_252.c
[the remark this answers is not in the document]
Enkelmann: Yes, it would make sense to add this as a comment to the file. I think you can simply copy-paste your explanation from here, it helped a lot in understanding the test cases.

src/cwe_checker_lib/src/abstract_domain/mod.rs
vobst: Good point, will change that.

test/src/lib.rs
Enkelmann: For the spurious function entries, `-O0` might again help. I know, you prefer testing on optimized code since it is more similar to real world code than `-O0` code. I think in the future, it would be better to have two sets of acceptance tests: One set of very simple test (like the current ones) that we test for each ISA-compiler pairing. Probably using `-O0` just to reduce the number of strange optimization edge cases we have to handle just for the CI. And one set of more complex tests that we only test for one or two ISA-compiler pairings (with `-O2`) where we can use knowledge about how the compiler will optimize the code to make the tests stable enough to run in the CI on each pull request.

For the difference in the extra layer of indirection in gcc we already talked a bit about it. Maybe there is no good solution, maybe we can change the check so that it reports the CWE always at the the end of the callee? I am content with blacklisting the test case for now (like you did). We could revisit the problem at a later time when we have some more experience on real-world binaries.

Enkelmann (changes requested):
Except for the `MemoryTaintMergeStrategy` everything looks fine. At least I think so, because I had to change my review process due to the force-push.

There is still the question on whether we should use `-O0` or `-O2` for the acceptance tests. In the end I will follow your decision here if you want to keep `-O2`. There also might be some old review comments that I have not yet marked as resolved. If some of them are resolved and I just missed it in the review, please tell me!

src/cwe_checker_lib/src/analysis/taint/state/memory_taint.rs
Enkelmann: But you defined the type `MemoryTaint` with `UnionMergeStrategy` instead of `MemoryTaintMergeStrategy`. So I think that the code does not use your intended implementation right now.

vobst:
> Except for the `MemoryTaintMergeStrategy` everything looks fine. 

This one should be fixed now.

> At least I think so, because I had to change my review process due to the force-push.

Yea, sorry, didn't expect it would throw GH off that much.

> 
> There is still the question on whether we should use `-O0` or `-O2` for the acceptance tests. In the end I will follow your decision here if you want to keep `-O2`. 

I've gone with O0 now since it means that more tests work (in the sense of reporting 9/9). I developed the tests with 02 in mind so I cannot tell if they still work as expected, only that the expected nr of warnings comes out in the end. I guess we will find out once we have more fine grained filtering that can also verify addresses and reasons.

I also modified one of the interprocedural test cases to cover the case where an object from the caller has to be merged with an offset. Seems to work as expected.

> There also might be some old review comments that I have not yet marked as resolved. If some of them are resolved and I just missed it in the review, please tell me!

I think all open conversions should be resolved now, but double check for yourself.

Have a nice weekend!

Enkelmann (approved):
Nothing left to nitpick for me, PR can be merged. Thanks for the great PR!

vobst:
> Nothing left to nitpick for me, PR can be merged. Thanks for the great PR!

Thanks for taking the time to review this PR and all of the explanations along the way!

I'd now rearrange the commits a bit to make them more suitable for the permanent commit log, i.e., write proper commit messages and combine/reorder things that belong together. This will not change the content of any commit. Then I'd also add one commit that updates the changelog for the current release in the end. Sounds OK?

vobst:
I have cleaned up the history and will merge it once the CI passes.

Is there an unwritten rule that the `CHANGES.md` must be onliners? I'd be adding the first multiline entry here :)

Enkelmann:
> Is there an unwritten rule that the `CHANGES.md` must be onliners? I'd be adding the first multiline entry here :)

No, that was just a habit. If you want to preserve the style, keep entries short (maybe 2 lines max?) and use the description texts of the referenced PRs for more detailed descriptions. But the style itself is only personal preference, you can change it up if you prefer a different style for the `CHANGES.md`.

vobst:
> > Is there an unwritten rule that the `CHANGES.md` must be onliners? I'd be adding the first multiline entry here :)
> 
> No, that was just a habit. If you want to preserve the style, keep entries short (maybe 2 lines max?) and use the description texts of the referenced PRs for more detailed descriptions. But the style itself is only personal preference, you can change it up if you prefer a different style for the `CHANGES.md`.

Okay, just wanted to be sure. It still renders as one line in the UI, looks like the limit there is about 120 chars and I tend to wrap at 80.

PR #452, opened 2024-04-02
WIP: test: add acceptance test for riscv64

vobst:
Compile all acceptance tests for riscv64 (clang and gcc).

# TODO:

- [ ] `register_calling_convention` is missing in the output of the Ghidra plugin (tested 10.2.3). This causes deserialization to fail on the Rust side.
- [ ] also add 32 bit RISCV; no gcc package available (double check); clang-10 should work (check)

### DIFF
--- a/test/lkm_samples/Dockerfile
+++ b/test/lkm_samples/Dockerfile
@@ -13,7 +13,6 @@ RUN set -x && \
     apt-get update && \
     apt-get install -y -q apt-utils dialog && \
     apt-get install -y -q \
-	aptitude \
 	bc \
 	bison \
 	bsdmainutils \


Enkelmann (commented):
Adding RISC-V to our test collection is a very good thing! :) However, I am inclined to suggest to postpone merging of this PR for the time being for two reasons:

First, we should also add 32-bit-RISC-V to the CI, because it might be more common in practice than the 64-bit variant.

Second, I do not think we should merge it as long as we cannot test it in the CI. It feels like adding dead code to the repository.

So my suggestion is to leave the PR open (maybe change it to WIP) for now and revisit it when we can actually add the test cases to the acceptance tests.

vobst:
This is now part of #472.

PR #454, opened 2024-04-02
Improve Docker image used for compiling artificial samples

vobst:
Some small improvements to the final image:
- smaller size (ignore artifacts)
- faster rebuild time (reordering of steps)
- better interactive usage (chown to user)


Enkelmann (approved):
Looks good to me, as long as the CI also agrees.

PR #455, opened 2024-04-02
Improve docker image size and add LKM config

vobst:
Drastically improves build time and image size for the `cwe_checker` Docker image. Also fixes an oversight I made when adding LKM support by including the new config in the container image.


Enkelmann (approved):
Looks good to me. But we should merge the master branch into this branch beforehand, so that the CI does not fail anymore

PR #456, opened 2024-04-02
docker/lkm_samples: do not install 'aptitude'

vobst:
We don't need it and it causes the CI to fail - two good reasons to get rid of it.


Enkelmann (approved):
LGTM

PR #462, opened 2024-05-03
Fix & improve Propagate Control Flow normalization pass

### LINKED ISSUE #461: panic in graph.rs because unwrap() of None
```
docker run --rm -e RUST_BACKTRACE=1 -v ./ip:/input ghcr.io/fkie-cad/cwe_checker /input
thread 'main' panicked at src/cwe_checker_lib/src/analysis/graph.rs:339:85:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0: rust_begin_unwind
             at ./rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/panicking.rs:645:5
   1: core::panicking::panic_fmt
             at ./rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/panicking.rs:72:14
   2: core::panicking::panic
             at ./rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/panicking.rs:144:5
   3: cwe_checker_lib::analysis::graph::GraphBuilder::add_jump_edge
   4: cwe_checker_lib::analysis::graph::get_program_cfg_with_logs
   5: cwe_checker::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

```
[ip.tar.gz](https://github.com/fkie-cad/cwe_checker/files/15060127/ip.tar.gz)

vobst:
Issue #461 discovered a problem with the Propagate Control Flow normalization pass: In rare cases it could happen that a basic block would be removed without re-targeting calls that return to it. [link](https://github.com/fkie-cad/cwe_checker/issues/461#issuecomment-2083642645)

This PR fixes the issue by allowing the re-targeting of call returns. Besides that, it includes a number of improvements to the optimization pass at large. In particular:

- [x] It adds support for the re-targeting of jumps without a known condition. Currently, a jump can only be re-targeted if some condition is known to be true whenever this branch is taken. This is necessary to resolve conditional branches, however, unconditional branches can be optimized away without a known condition.
- [x] It adds support for the re-targeting of call returns. This is what fixes #461. Note that we cannot assume that conditions remain true across calls due to possible side-effects so the first change is required to make this work.
- [x] It adds support to derive the block precondition from multiple incoming edges. Currently, a block precondition can only be derived if the block has a single incoming edge that stems from a conditional jump. However, we can still derive a precondition if there are multiple incoming edges that have the same condition.
- [x] It adds support to remember precondition and branch condition when re-targeting conditional jumps. Currently, we only remember the branch condition, but it turns out that there are cases where computing and remembering the precondition helps as well.
- [x] It updates the docs and makes minor code-style modernizations.

# Testing

There is a unit test for each of the proposed changes that make the optimization more aggressive. Furthermore, the optimizations of the new pass were manually verified (at random) for three programs for three different architectures (MIPS, ARM32, AMD64).

# Measurements

The following table includes a comparison of this PR and the current optimization pass. It show the number of basic block nodes and jump edges in the CFG of the unoptimized `readelf` IR program for different host architectures. Next to that are the fractions of basic blocks and jump edges that can be removed by the current and proposed optimization passes respectively. In short, the improved pass can remove ~1.8x more stuff from the CFG than the old one.

One interesting bit is that the optimization performs performs worst for MIPS, which was the architecture for which it was originally introduced.

```
|             | unoptimized           | master                   | PR                                                |
|             | BB        | Jumps     | dBB [%]     | dJumps [%] | dBB [%] | dJumps  [%] | impr. BB    | impr. Jumps |
|-------------|-----------|-----------|-------------|------------|---------|-------------|-------------|-------------|
| arm64       | 35,462.00 | 40,296.00 | 2.35        | 2.07       | 6.20    | 5.46        | 2.64        | 2.63        |
| armel       | 50,323.00 | 60,404.00 | 3.26        | 4.01       | 6.77    | 6.93        | 2.08        | 1.73        |
| armhf       | 50,385.00 | 61,036.00 | 3.58        | 4.49       | 6.89    | 7.22        | 1.92        | 1.61        |
| mipsel      | 27,784.00 | 31,661.00 | 0.63        | 0.58       | 0.69    | 0.64        | 1.09        | 1.09        |
| amd64       | 48,556.00 | 55,543.00 | 3.20        | 2.80       | 5.48    | 4.84        | 1.71        | 1.73        |
|-------------|-----------|-----------|-------------|------------|---------|-------------|-------------|-------------|
| avrg.       | 42,502.00 | 49,788.00 | 2.60        | 2.79       | 5.21    | 5.02        | 1.89        | 1.76        |
| std. dev.   | 10,321.60 | 13,143.66 | 1.19        | 1.56       | 2.58    | 2.64        | 0.56        | 0.56        |

```


src/cwe_checker_lib/src/analysis/graph.rs
```diff
@@ -531,6 +531,73 @@ pub fn get_entry_nodes_of_subs(graph: &Graph) -> HashMap<Tid, NodeIndex> {
...
+                Node::CallReturn { .. } => call_return_nodes += 1,
+                Node::CallSource { .. } => call_source_nodes += 1,
+            }
+        }
+
+        nodes_map.insert("total".into(), total_nodes.into());
+        nodes_map.insert("blk_start".into(), blk_start_nodes.into());
+        nodes_map.insert("blk_end".into(), blk_end_nodes.into());
+        nodes_map.insert("call_return".into(), call_return_nodes.into());
+        nodes_map.insert("call_source".into(), call_source_nodes.into());
```
Enkelmann: I suppose you used this code for computing your statistics? I would not use the `ToJsonCompact` trait for that, because I would expect it to print an actual representation of the graph, where one could look up if e.g. a specific node has the expected incoming and outgoing edges. But it is only my preference, you might disagree with me here.

src/cwe_checker_lib/src/intermediate_representation/project/propagate_control_flow.rs
```diff
@@ -247,18 +351,21 @@ fn negate_condition(expr: Expression) -> Expression {
     }
 }
 
-/// Iterates the CFG and returns all node's blocks, that do not have an incoming edge.
+/// Iterates the CFG and returns all node that do not have an incoming edge.
```
Enkelmann: Typo `nodes` instead of `node`

src/cwe_checker_lib/src/intermediate_representation/project/propagate_control_flow.rs
```diff
@@ -280,10 +387,15 @@ pub mod tests {
     use crate::{def, expr};
     use std::collections::BTreeMap;
 
-    fn mock_condition_block(name: &str, if_target: &str, else_target: &str) -> Term<Blk> {
+    fn __mock_condition_block(
```
Enkelmann: This should have another name. Especially, because starting it with an underscore suggests that the function is not used.

Enkelmann (changes requested):
Except for some very small nitpicks, looks good to me.

By the way: The reason that it performs so bad on MIPS is probably because we still have a known error when parsing conditional assignment instructions in MIPS. Since the whole control flow path propagation exists mainly for improving the control flow in the presence of conditional assignments, it cannot do much if the instructions are not parsed correctly.

**Edit:** The bug fix might not be complete, see my other comment. You can probably check with some elaborated unit test whether I am right or not.

Enkelmann:
I might have an idea why the old code did not catch the case: We use the CFG to check for incoming edges when removing blocks, right? But if a callee has no return instruction, then the CFG may not contain an edge to the return site. But the TID of the return site is nevertheless referenced in the call instruction. Your code solves that by retargeting the return site. In theory it might be possible that the return site cannot be retargeted, while all other edges to the return site are retargeted nevertheless. So I think the bug still persists, it is just less likely now...

vobst:
> I might have an idea why the old code did not catch the case: We use the CFG to check for incoming edges when removing blocks, right? But if a callee has no return instruction, then the CFG may not contain an edge to the return site. But the TID of the return site is nevertheless referenced in the call instruction. Your code solves that by retargeting the return site. In theory it might be possible that the return site cannot be retargeted, while all other edges to the return site are retargeted nevertheless. So I think the bug still persists, it is just less likely now...

Good catch! It is indeed not too difficult to construct this situation. Essentially a case where a call to a non-returning function is not recognized as such and "returns" to a conditional block that can be optimized away. See below for a concrete example.

I see a couple of options how to proceed:

a.) Add another, preceding normalization pass that recognizes calls that have `Some` return TID and for which no artificial return nodes are generated in the CFG for some reason. Set the return TID to `None` for those calls, essentially marking them non-returning.
b.) Keep track of all call-returns manually in the propagate CF pass and make sure that their target is only removed if they are re-targeted as well.
c.) It is already quite unlikely to hit the original case (at least it took us quite a while to notice) and will probably take much longer for a binary to emerge that triggers the remaining edge case. So just ignore it.

b.) and c.) are both unsatisfactory. For a.) it would be interesting to know if you have a feeling whether there are any bad surprises waiting down the line. On the first sight it seems like the return target information is worthless if the artificial return nodes are not generated so we might as well throw it away entirely (Indirect calls and call-other are handled differently and are not affected, right?). The question is rather why Ghidra generated it in the first place.

```
  sub_1                                                 sub_2           
 ┌───────────────────────────────────────────────┐      ┌─────────────┐ 
 │                                               │      │             │ 
 │                               ┌────────────┐  │      │ ┌────────┐  │ 
 │                               │            │  │      │ │        │  │ 
 │                 ┌─────┐       │ call sub_2 │  │      │ │  loop  │  │ 
 │                 │     │       │            │  │      │ │        │  │ 
 │                 │  C  │       └────────────┘  │      │ └───┬───▲┘  │ 
 │                 │     │                       │      │     │   │   │ 
 │                 └┬───┬┘              │ return │      │     └───┘   │ 
 │                  │   │               │        │      │             │ 
 │                  │   │                        │      └─────────────┘ 
 │                  │   │            ┌─────┐     │                      
 │                  │   │   C        │     │     │                      
 │                  │   └───────────►│  C  │     │                      
 │ ┌─────┐ not C    │                │     │     │                      
 │ │     ◄──────────┘                └─┬─┬─┘     │                      
 │ │ E_1 │               not C         │ │       │                      
 │ │     ◄─────────────────────────────┘ │       │                      
 │ └─────┘                               │ C     │                      
 │                                       │       │                      
 │                                   ┌───▼───┐   │                      
 │                                   │       │   │                      
 │                                   │  E_2  │   │                      
 │                                   │       │   │                      
 │                                   └───────┘   │                      
 │                                               │                      
 └───────────────────────────────────────────────┘                      
```

In this example the first `C` branch can be re-targeted which means that the second conditional block becomes orphaned as the call does not cause an incoming edge due to the missing return statement.

This unit test uses the above example to show that we can still trigger this bug under those conditions.

```rust
#[test]
fn call_return_to_cond_jump_removed() {
    let sub_1 = Sub {
        name: "sub_1".to_string(),
        calling_convention: None,
        blocks: vec![
            mock_condition_block("cond_blk_1", "cond_blk_2", "end_blk_1"),
            mock_block_with_defs_and_call("call_blk", "sub_2", "cond_blk_2"),
            mock_condition_block("cond_blk_2", "end_blk_2", "end_blk_1"),
            mock_block_with_defs("end_blk_1", "end_blk_1"),
            mock_block_with_defs("end_blk_2", "end_blk_2"),
        ],
    };
    let sub_1 = Term {
        tid: Tid::new("sub_1"),
        term: sub_1,
    };
    let sub_2 = Sub {
        name: "sub_2".to_string(),
        calling_convention: None,
        blocks: vec![mock_block_with_defs("loop_block", "loop_block")],
    };
    let sub_2 = Term {
        tid: Tid::new("sub_2"),
        term: sub_2,
    };
    let mut project = Project::mock_arm32();
    project.program.term.subs =
        BTreeMap::from([(Tid::new("sub_1"), sub_1), (Tid::new("sub_2"), sub_2)]);

    let cfg_before_normalization = graph::get_program_cfg(&project.program);
    cfg_before_normalization.print_compact_json();

    propagate_control_flow(&mut project);
    // construction of CFG would panic now
    //graph::get_program_cfg(&project.program);
    let expected_blocks = vec![
        // `cond_blk_1` can be retarget.
        mock_condition_block("cond_blk_1", "end_blk_2", "end_blk_1"),
        // `call_blk` can not be re-targetd since no condition is known.
        mock_block_with_defs_and_call("call_blk", "sub_2", "cond_blk_2"),
        // `cond_blk_2` was removed since cond_blk_1 was re-targeted.
        // Note: `call_blk` did not contribute an incoming edge since the
        // callee does not return.
        mock_block_with_defs("end_blk_1", "end_blk_1"),
        mock_block_with_defs("end_blk_2", "end_blk_2"),
    ];

    assert_eq!(
        &project.program.term.subs[&Tid::new("sub_1")].term.blocks[..],
        &expected_blocks[..]
    );
}
```

PS: How do you like the updated `ToJsonCompact` impl ;)
```
{
  "edge_counts": {
    "block": 6,
    "call": 1,
    "call_combine": 1,
    "cr_call_stub": 0,
    "cr_return_stub": 0,
    "extern_call_stub": 0,
    "jump": 7,
    "return_combine": 0,
    "total": 15
  },
  "edges": {
    "0 -> 1": "Block",
    "1 -> 4": "Jump",
    "1 -> 6": "Jump",
    "10 -> 11": "Block",
    "11 -> 10": "Jump",
    "12 -> 10": "Call",
    "2 -> 3": "Block",
    "3 -> 12": "CallCombine",
    "4 -> 5": "Block",
    "5 -> 6": "Jump",
    "5 -> 8": "Jump",
    "6 -> 7": "Block",
    "7 -> 6": "Jump",
    "8 -> 9": "Block",
    "9 -> 8": "Jump"
  },
  "node_counts": {
    "blk_end": 6,
    "blk_start": 6,
    "call_return": 0,
    "call_source": 1,
    "total": 13
  },
  "nodes": {
    "0": "BlkStart @ cond_blk_1 (sub sub_1)",
    "1": "BlkEnd @ cond_blk_1 (sub sub_1)",
    "10": "BlkStart @ loop_block (sub sub_2)",
    "11": "BlkEnd @ loop_block (sub sub_2)",
    "12": "CallSource @ loop_block (sub sub_2) (caller @ call_blk (sub sub_1))",
    "2": "BlkStart @ call_blk (sub sub_1)",
    "3": "BlkEnd @ call_blk (sub sub_1)",
    "4": "BlkStart @ cond_blk_2 (sub sub_1)",
    "5": "BlkEnd @ cond_blk_2 (sub sub_1)",
    "6": "BlkStart @ end_blk_1 (sub sub_1)",
    "7": "BlkEnd @ end_blk_1 (sub sub_1)",
    "8": "BlkStart @ end_blk_2 (sub sub_1)",
    "9": "BlkEnd @ end_blk_2 (sub sub_1)"
  }
}
```

src/cwe_checker_lib/src/analysis/graph.rs
vobst: Yea, agreed, I was just being lazy and thought I'd add that functionality when I need it. Turns out, I needed it right now to debug the other issue you found :D

Enkelmann:
Your updated `ToJsonCompact` format looks very good!

Right now, your option a) would probably result in the best CFG that we can generate. Just make sure to mention it prominently in the doc-comment of the normalization pass. Maybe also add a INFO-level log message for each function (or call?) deemed non-returning. Just to make it easier to spot real-world binaries where suspiciously many functions are marked as non-returning.

vobst:
> Your updated `ToJsonCompact` format looks very good!
> 

Nice, I included that in the first new commit.

> Right now, your option a) would probably result in the best CFG that we can generate. Just make sure to mention it prominently in the doc-comment of the normalization pass. Maybe also add a INFO-level log message for each function (or call?) deemed non-returning. Just to make it easier to spot real-world binaries where suspiciously many functions are marked as non-returning.

I decided to split up the existing `remove_references_to_nonexisting_tids_and_retarget_non_returning_calls` pass into three passes:

1. Unconditionally add the artificial sinks. (As far as I can see there is no problem with having them when they are not needed and it simplifies the code.)
2. `remove_references_to_nonexisting_tids`: The part of the code in `remove_references_to_nonexisting_tids_and_retarget_non_returning_calls` that was responsible for that; it can now assume that artificial sinks always exist.
3. `retarget_non_returning_calls_to_artifical_sink`: handles calls to external functions AND functions without a return instruction; it can now assume that artificial sinks always exist.

Here are some stats about how often this new pass strikes. All in all they suggest that it is a pretty rare thing. I included some reasons for false positives in the doc comment.

```
| $arch-$binary        | functions | non-returning | calls | non-returning |
|----------------------|-----------|---------------|-------|---------------|
| arm64-readelf        | 429       | 0             | 8497  | 0             |
| armel-readelf        | 365       | 0             | 8046  | 0             |
| armhf-readelf        | 369       | 0             | 7939  | 0             |
| mipsel-readelf       | 346       | 9             | 5213  | 27            |
| amd64-readelf        | 681       | 7             | 10094 | 0             |
| armel-ls             | 235       | 0             | 1190  | 0             |
| mipsel-ls            | 220       | 7             | 1409  | 14            |
| arm64-ls             | 348       | 0             | 1551  | 0             |
| x86-ls               | 338       | 10            | 1460  | 134           |
| armhf-ls             | 162       | 5             | 731   | 3             |
| amd64-ls             | 322       | 4             | 1400  | 0             |
| ppc64el-ls           | 435       | 1             | 2443  | 0             |
| x86-netfs.ko         | 37        | 6             | 79    | 0             |
| powerpc64le-netfs.ko | 109       | 10            | 364   | 0             |
| mips64r2el-netfs.ko  | 52        | 0             | 720   | 0             |
| mips32r2el-netfs.ko  | 54        | 0             | 790   | 0             |
| amd64-netfs.ko       | 62        | 0             | 310   | 0             |
| arm64-netfs.ko       | 55        | 0             | 194   | 0             |
| armhf-netfs.ko       | 54        | 0             | 227   | 0             |
```

However, these stats may also be a bit biased. On the IoT `ip` executable from the original issue we get:

```
total_fn = 366
total_non_ret_fn = 11
total_calls_with_ret = 3469
total_retargeted_calls = 414
```

Maybe something with the compiler+settings. In general it does no harm to retarget those calls as they can not be analyzed anyway - however - maybe we can address some of the root causes an another point.

src/cwe_checker_lib/src/intermediate_representation/project/propagate_control_flow.rs
vobst: Thanks, fixed that.

src/cwe_checker_lib/src/intermediate_representation/project/propagate_control_flow.rs
vobst: mumbles something about being lazy when coding late at night ... but changed that.

src/cwe_checker_lib/src/intermediate_representation/project.rs
```diff
@@ -288,10 +348,14 @@ impl Project {
     #[must_use]
     pub fn normalize(&mut self) -> Vec<LogMessage> {
         let mut logs = self.remove_duplicate_tids();
+        self.add_artifical_sinks();
+        logs.append(self.remove_references_to_nonexisting_tids().as_mut());
+        make_block_to_sub_mapping_unique(self);
         logs.append(
-            &mut self.remove_references_to_nonexisting_tids_and_retarget_non_returning_calls(),
+            self.retarget_non_returning_calls_to_artificial_sink()
```
vobst, line 377: Just noticed that running this pass after `make_block_to_sub_mapping_unique` means that the Artificial Sink Block that the calls "return" to will be shared. However, running the pass beforehand eliminates some cases where functions would be misclassified as non returning. Maybe run it again afterwards?

Enkelmann: Yes, running it again would be a viable solution. And we need to make the block-to-sub-mapping unique again, as this invariant is used later on.

Enkelmann (approved):
Apart from the issue you mentioned yourself, everything looks good to me.

src/cwe_checker_lib/src/intermediate_representation/project.rs
vobst: There is an annoying corner case where it might happen that a Sub already has an artificial sink block (due to an earlier pass) and running non-ret-call-return-retargeting+inlining would add a second one.

I now modified the non-ret-call-return-retargeting pass s.t. it preserves a unique block-to-sub mapping. It now essentially maps them directly to the Sub's artificial sink without the detour through the extra pass.

vobst:
> Apart from the issue you mentioned yourself, everything looks good to me.

Thanks for the review! Then I'll now clean up the commit log and resolve the merge conflicts with the benchmarking PR.

src/cwe_checker_lib/src/intermediate_representation/project.rs
```diff
@@ -109,107 +108,198 @@ impl Project {
...
+                }
+            })
+            .collect();
+
+        (non_returning_subs, log_messages)
+    }
+
+    /// Adds a function that serves as an artificial sink in the CFG.
+    fn add_artifical_sinks(&mut self) {
+        let artificial_sink_sub: Term<Sub> = Term {
```
vobst: make this a method on `Term<Sub>`.

src/cwe_checker_lib/src/intermediate_representation/project.rs
```diff
@@ -109,107 +108,198 @@ impl Project {
...
                     log_messages.append(&mut logs);
                 }
                 for jmp in block.term.jmps.iter_mut() {
-                    if let Err(log_msg) = jmp.retarget_nonexisting_jump_targets_to_dummy_tid(
-                        &jump_target_tids,
-                        &dummy_sub_tid,
-                        &dummy_blk_tid,
-                    ) {
+                    if let Err(log_msg) =
+                        jmp.retarget_nonexisting_jump_targets_to_dummy_tid(&all_jump_targets)
```
vobst: rename `..._to_artificial_sink`.

src/cwe_checker_lib/src/intermediate_representation/term.rs
```diff
@@ -14,12 +14,25 @@ pub struct Tid {
 }
 
 impl Tid {
+    /// Prefix for IDs of artificial sinks in the control flow graph.
+    ///
+    /// Dummy blocks with such TIDs are added for different purposes, e.g., as
+    /// targets for jumps to non-existing targets or return targets for calls to
+    /// non-returning functions
```
vobst: . missing

src/cwe_checker_lib/src/intermediate_representation/project/propagate_control_flow.rs
```diff
@@ -1,75 +1,134 @@
...
+                        return_: Some(return_target),
+                    },
+            }] => {
+                if let Some(new_target) = find_target_for_retargetable_jump(
+                    return_target,
+                    &sub.term,
+                    // Call may have side-effects that invalidate our
+                    // knowledge about any condition we know to be true
+                    // after execution of all DEFs in a block.
+                    &Vec::new(),
```
vobst: `with_capacity(0)`

src/cwe_checker_lib/src/intermediate_representation/project/propagate_control_flow.rs
```diff
@@ -1,75 +1,134 @@
...
-                            jmps_to_retarget.insert(jump_tid.clone(), new_target);
-                        }
-                    }
+    for node in cfg_before_normalization.node_indices() {
+        let Node::BlkStart(block, sub) = cfg_before_normalization[node] else {
+            continue;
+        };
+        // Conditions that we know to be true "on" a particular outgoing
+        // edge.
+        let mut true_conditions = Vec::new();
```
vobst: `with_capacity(2)`

vobst (commented):
Just a few small nits.

vobst (commented):
Looks like force push did not break anything.

